PR #10, opened 2020-10-09
adding support of eu-north-1(Stockholm) region

rvoitenko:
* AWS Lambda Extensions API is also available in eu-north-1(Stockholm) region, so it would be nice to have official Hashicorp's Lambda layer presented there as well


tomhjp (approved):
Thanks for raising this! I've pushed a few versions to get `eu-north-1` up to the same revision number as the other regions, so you should now be able to use `arn:aws:lambda:eu-north-1:634166935893:layer:vault-lambda-extension:6`. That ARN worked fine in my testing, although I noticed `t2` instances are not available in `eu-north-1`; I'll add a note about using `t3` instead.

PR #15, opened 2020-11-10
Fix/quick start

burtlo:
Fixes RDS requirement of VPC/Subnet/Gateway

- RDS wanted two subnets
- Expanded the VPC
- Created two subnets
- Internet Gateway was required
- Adds the Vault server to Subnet

Currently failing reaching the Vault server. It's not reachable at all.

```shell
aws_instance.vault-server (remote-exec): Connecting to remote host via SSH...
aws_instance.vault-server (remote-exec):   Host: 54.146.237.94
aws_instance.vault-server (remote-exec):   User: ubuntu
aws_instance.vault-server (remote-exec):   Password: false
aws_instance.vault-server (remote-exec):   Private key: true
aws_instance.vault-server (remote-exec):   Certificate: false
aws_instance.vault-server (remote-exec):   SSH Agent: true
aws_instance.vault-server (remote-exec):   Checking Host Key: false
aws_instance.vault-server: Still creating... [2m30s elapsed]
aws_instance.vault-server: Still creating... [2m40s elapsed]
aws_instance.vault-server: Still creating... [2m50s elapsed]
```

### DIFF
--- a/quick-start/demo-function/main.go
+++ b/quick-start/demo-function/main.go
@@ -44,10 +44,10 @@ func HandleRequest(ctx context.Context, payload Payload) error {
 	}
 
 	// read token
-	tokenRaw, err := ioutil.ReadFile("/tmp/vault/token")
-	if err != nil {
-		return fmt.Errorf("error reading file: %w", err)
-	}
+	// tokenRaw, err := ioutil.ReadFile("/tmp/vault/token")
+	// if err != nil {
+	// 	return fmt.Errorf("error reading file: %w", err)
+	// }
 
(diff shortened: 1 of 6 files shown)


tomhjp:
Thanks for looking into this. After working on this offline, we discovered that the issues were to do with running on an account pre-2013-12-04 which has no default VPC ([docs](https://docs.aws.amazon.com/vpc/latest/userguide/default-vpc.html)). Going to close this for now in favour of just using more recent accounts, but I'm still open to an example that doesn't rely on the default VPC if anyone wants to work on it.

PR #17, opened 2020-11-12
Default Vault Auth Role To Lambda Name

abrucker235:
Changed it so that consumers could create auth roles based on lambda name and don't have to set additional variable unless they want it to be called different.

### DIFF
--- a/main.go
+++ b/main.go
@@ -51,7 +51,12 @@ func initialiseExtension(logger *log.Logger) {
 	logger.Println("Initialising")
 
 	vaultAddr := os.Getenv("VAULT_ADDR")
+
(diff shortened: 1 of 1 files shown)


catsby:
Hey @abrucker235 - thanks for the contribution here! I admit I'm reluctant to merge this however without knowing if this is a common pattern for our users, or if it's particular to your specific usage patterns. I hope that makes sense.

abrucker235:
@catsby this might be a pattern we use and I can see the hesitation, but its a pattern I feel helps encourage not reusing vault roles across different lambdas and practice least access.  Technically this doesn't turn off the original capability.

PR #22, opened 2020-12-10
Add note about supported runtimes

tomhjp:
This is an easy to miss gotcha when using the `go1.x` runtime which does not support extensions. If the layer is added for an unsupported runtime, the extension won't be invoked at all, so it's pretty hard to debug.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Requirements:
 * An instance of Vault accessible from AWS Lambda
 * An authenticated `vault` client
 * A secret in Vault that you want your Lambda to access, and a policy giving read access to it
-* **Your Lambda must run in one of the [lambda-supported-runtimes] for extensions**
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,6 +37,7 @@ Requirements:
 * An instance of Vault accessible from AWS Lambda
 * An authenticated `vault` client
 * A secret in Vault that you want your Lambda to access, and a policy giving read access to it
+* **Your Lambda must use one of the [supported runtimes][lambda-supported-runtimes] for extensions**
```
catsby, line 40:
```suggestion
* **Your Lambda function must use one of the [supported runtimes][lambda-supported-runtimes] for extensions**
```

catsby (approved):
LGTM, accept minor nit at your discretion

Issue #13, opened 2020-10-27
Passing of secrets not in line with best practice


ineffyble:
This extension only allows the user to write the secrets to a file in the filesystem (e.g. `/tmp/...`) as a way to pass secrets to the Lambda function that needs them.

This is against [AWS best practice](https://docs.aws.amazon.com/lambda/latest/dg/best-practices.html), which states "_don’t use the execution environment to store user data, events, or other information with security implications_".

Looking at the Lambda Extensions API, it should be possible for the Vault extension to set an environment variable instead, which can be read by the Lambda process.

tomhjp:
Hi, thanks for opening this!

You're absolutely right that it's not best practice, and it's one of the things we wanted to get feedback on during the public preview period. Unfortunately, because this is an external extension which runs as its own process next to the function (as opposed to an internal extension), I don't think we can set environment variables for the main function process.

What we could do though is create a server mode that listens on a local socket with some API for requesting secrets. Then the extension wouldn't do much at startup, and instead respond to requests during execution time, passing secrets to the function through a TCP or unix socket. This is something we've discussed adding to the extension, but we don't have a firm design or timeframe yet - would it solve your concern? It's obviously a little more complex to use as a consumer, but hopefully that trade-off would be worth it to lots of users. On that note, I would love to understand more about your use case and how this problem impacts you if you're happy to share! 🙂

ineffyble:
@tomhjp That sounds great - would certainly address my concern.

My context is working in a financial services environment with strict security requirements and controls. We're just starting to migrate to Hashicorp Vault for secrets management, and the extension seemed like a great way to accomplish this for Serverless microservices.

AndreZiviani:
Hi @tomhjp 

I have a different use case that would benefit from this as well, I would like to dynamically access a Vault secret based on a parameter from the request, this TCP/socket interface would simplify my lambda a lot!

dmattia:
+1 for the server idea. In ECS/EKS/EC2s where we use Vault in AWS, we can use the `vault agent`. I would love it if this extension worked similarly, allowing us to communicate with it just like the agent endpoint.

tomhjp:
Proxy server mode released in layer version 8/git tag v0.3.0 🎉 

Check out the demo function in `quick-start` for a concrete example of consuming the proxy from Lambda, and please feel free to raise further issues with feedback! Thanks all for your input.

PR #27, opened 2021-02-24
Add server mode

tomhjp:
Addresses #13.

* Start local HTTP server at 127.0.0.1:8200 with proxy handler
  * The proxy will add authentication to requests based on the IAM auth configured via environment variables
* Add local integration test with docker-compose
* No longer require >=1 secrets configured via env var

To run the integration test on your machine, set the following env vars:
`AWS_ROLE_ARN` as the role you want IAM auth to identify you as, e.g. `arn:aws:iam::<account number>:role/vault_team_dev-developer`
`AWS_SECRET_ACCESS_KEY`, `AWS_ACCESS_KEY_ID` and `AWS_SESSION_TOKEN` with some AWS credentials that can access the above role.

Then from the root of the repo:

```bash
docker-compose --file test/docker-compose.yaml up --abort-on-container-exit --exit-code-from lambda
```

Still TODO:
- [x] Finalise error codes/error handling in the proxy handler
- [x] More unit test cases
- [x] Documentation updates
- [x] Finish changelog entry

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,9 @@
 # vault-lambda-extension
 
 This repository contains the source code for HashiCorp's Vault AWS Lambda extension.
-The extension utilizes the AWS Lambda Extensions API to read secrets from your
-Vault deployment and write the result to disk before the Lambda function itself
-starts to execute. To use it, include the following ARN as a layer in your
-Lambda function:
+The extension utilizes the AWS Lambda Extensions API to help your Lambda function
+read secrets from your Vault deployment. To use it, include the following ARN as a
+layer in your Lambda function:
 
 ```text
 arn:aws:lambda:us-east-1:634166935893:layer:vault-lambda-extension:7
@@ -16,8 +15,14 @@ Where region may be any of `ap-northeast-1`, `ap-northeast-2`, `ap-south-1`,
 `us-west-1`, `us-west-2`.
 
 The extension authenticates with Vault using [AWS IAM auth][vault-aws-iam-auth],
-and writes the result as JSON to disk. It also writes a vault token to
-`/tmp/vault/token`. All configuration is supplied via environment variables.
+and all configuration is supplied via environment variables. There are two methods
+to read secrets, which can both be used side-by-side:
+
+* Configure environment variables such as `VAULT_SECRET_PATH` for the extension
+  to read a secret and write it to disk.
+* Make unauthenticated requests to the extension's local server at
+  `http://127.0.0.1:8200`, which will add an authentication header and proxy to
+  the configured `VAULT_ADDR`.
 
 ## Getting Started
 
@@ -72,18 +77,23 @@ material from `/tmp/vault/secret.json`. The exact contents of the JSON object
 will depend on the secret read, but its schema is the [Secret struct][vault-api-secret-struct]
 from the Vault API module.
 
+Alternatively, you can send normal Vault API requests over HTTP to the local
+proxy at `http://127.0.0.1:8200`, and the extension will add authentication
+before forwarding the request. Vault responses will be returned unmodified.
+
 ## Configuration
 
 The extension is configured via [Lambda environment variables][lambda-env-vars].
 Most of the [Vault CLI client's environment variables][vault-env-vars] are available,
 as well as some additional variables to configure auth, which secret(s) to read and
-where to write secrets. At least one valid secret to read must be specified.
+where to write secrets.
 
 Environment variable    | Description | Required | Example value
 ------------------------|-------------|----------|--------------
 `VAULT_ADDR`            | Vault address to connect to | Yes | `https://x.x.x.x:8200`
 `VAULT_AUTH_PROVIDER`   | Name of the configured AWS IAM auth route on Vault | Yes | `aws`
 `VAULT_AUTH_ROLE`       | Vault role to authenticate as | Yes | `lambda-app`
+`VAULT_IAM_SERVER_ID`   | Value to pass to the Vault server via the [`X-Vault-AWS-IAM-Server-ID` HTTP Header for AWS Authentication](https://www.vaultproject.io/api-docs/auth/aws#iam_server_id_header_value) | No | `vault.example.com`
 `VAULT_SECRET_PATH`     | Secret path to read, written to `/tmp/vault/secret.json` unless `VAULT_SECRET_FILE` is specified | No | `database/creds/lambda-app`
 `VAULT_SECRET_FILE`     | Path to write the JSON response for `VAULT_SECRET_PATH` | No | `/tmp/db.json`
 `VAULT_SECRET_PATH_FOO` | Additional secret path to read, where FOO can be any name, as long as a matching `VAULT_SECRET_FILE_FOO` is specified | No | `secret/lambda-app/token`
@@ -101,7 +111,6 @@ Environment variable    | Description | Required | Example value
 `VAULT_CLIENT_CERT`     | Path to a PEM-encoded client certificate on the local disk | No | `/tmp/client.crt`
 `VAULT_CLIENT_KEY`      | Path to an unencrypted, PEM-encoded private key on disk which corresponds to the matching client certificate | No | `/tmp/client.key`
 `VAULT_CLIENT_TIMEOUT`  | Timeout for Vault requests. Default value is 60s. Ignored by proxy server. **Any value over 10s will exceed the Extensions API timeout and therefore have no effect** | No | `5s`
-`VAULT_IAM_SERVER_ID`   | Value to pass to the Vault server via the [`X-Vault-AWS-IAM-Server-ID` HTTP Header for AWS Authentication](https://www.vaultproject.io/api-docs/auth/aws#iam_server_id_header_value) | No | `vault.example.com`
 `VAULT_MAX_RETRIES`     | Maximum number of retries on `5xx` error codes. Defaults to 2. Ignored by proxy server | No | `2`
 `VAULT_SKIP_VERIFY`     | Do not verify Vault's presented certificate before communicating with it. Setting this variable is not recommended and voids Vault's [security model][vault-security-model]  | No | `true`
 `VAULT_TLS_SERVER_NAME` | Name to use as the SNI host when connecting via TLS | No | `vault.example.com`
--- a/README.md
+++ b/README.md
@@ -100,15 +100,13 @@ Environment variable    | Description | Required | Example value
 `VAULT_CAPATH`          | Path to a _directory_ of PEM-encoded CA certificate files on the local disk | No | `/tmp/certs`
 `VAULT_CLIENT_CERT`     | Path to a PEM-encoded client certificate on the local disk | No | `/tmp/client.crt`
 `VAULT_CLIENT_KEY`      | Path to an unencrypted, PEM-encoded private key on disk which corresponds to the matching client certificate | No | `/tmp/client.key`
-`VAULT_CLIENT_TIMEOUT`  | Timeout for Vault requests. Default value is 60s. **Any value over 10s will exceed the Extensions API timeout and therefore have no effect** | No | `5s`
+`VAULT_CLIENT_TIMEOUT`  | Timeout for Vault requests. Default value is 60s. Ignored by proxy server. **Any value over 10s will exceed the Extensions API timeout and therefore have no effect** | No | `5s`
 `VAULT_IAM_SERVER_ID`   | Value to pass to the Vault server via the [`X-Vault-AWS-IAM-Server-ID` HTTP Header for AWS Authentication](https://www.vaultproject.io/api-docs/auth/aws#iam_server_id_header_value) | No | `vault.example.com`
-`VAULT_MAX_RETRIES`     | Maximum number of retries on `5xx` error codes. Defaults to 2 | No | `2`
+`VAULT_MAX_RETRIES`     | Maximum number of retries on `5xx` error codes. Defaults to 2. Ignored by proxy server | No | `2`
 `VAULT_SKIP_VERIFY`     | Do not verify Vault's presented certificate before communicating with it. Setting this variable is not recommended and voids Vault's [security model][vault-security-model]  | No | `true`
 `VAULT_TLS_SERVER_NAME` | Name to use as the SNI host when connecting via TLS | No | `vault.example.com`
-`VAULT_RATE_LIMIT`      | Only applies to a single invocation of the extension. See [Vault Commands (CLI)][vault-env-vars] documentation for details | No | `10`
-`VAULT_NAMESPACE`       | The namespace to use for the command | No | `education`
-`VAULT_SRV_LOOKUP`      | The Vault client will lookup DNS SRV records for the host. See [Vault Commands (CLI)][vault-env-vars] documentation for details | No | `true`
-`VAULT_MFA`             | MFA credentials. See [Vault Commands (CLI)][vault-env-vars] documentation for details | No | `true`
+`VAULT_RATE_LIMIT`      | Only applies to a single invocation of the extension. See [Vault Commands (CLI)][vault-env-vars] documentation for details. Ignored by proxy server | No | `10`
+`VAULT_NAMESPACE`       | The namespace to use for pre-configured secrets. Ignored by proxy server | No | `education`
 
 ## Limitations
 
--- a/main.go
+++ b/main.go
@@ -85,9 +85,9 @@ func initialiseExtension(logger *log.Logger, threadFinishedCh chan struct{}) (*h
 
 	client, err := vault.NewClient(logger, vaultAuthRole, vaultAuthProvider, vaultIAMServerID)
 	if err != nil {
-		return nil, fmt.Errorf("error getting client: %s", err)
+		return nil, fmt.Errorf("error getting client: %w", err)
 	} else if client == nil {
-		return nil, fmt.Errorf("nil client returned: %s", err)
+		return nil, fmt.Errorf("nil client returned: %w", err)
 	}
 
 	err = writePreconfiguredSecrets(client)
@@ -118,7 +118,7 @@ func initialiseExtension(logger *log.Logger, threadFinishedCh chan struct{}) (*h
 func writePreconfiguredSecrets(client *api.Client) error {
 	configuredSecrets, err := config.ParseConfiguredSecrets()
 	if err != nil {
-		return fmt.Errorf("failed to parse configured secrets to read: %s", err)
+		return fmt.Errorf("failed to parse configured secrets to read: %w", err)
 	}
 
 	if _, err := os.Stat(config.DefaultSecretDirectory); os.IsNotExist(err) {
@@ -137,7 +137,7 @@ func writePreconfiguredSecrets(client *api.Client) error {
 		// Will block until shutdown event is received or cancelled via the context.
 		secret, err := client.Logical().Read(s.VaultPath)
 		if err != nil {
-			return fmt.Errorf("error reading secret: %s", err)
+			return fmt.Errorf("error reading secret: %w", err)
 		}
 
 		content, err := json.MarshalIndent(secret, "", "  ")
--- a/main.go
+++ b/main.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/signal"
 	"path"
+	"sync"
 	"syscall"
 
 	"github.com/hashicorp/vault-lambda-extension/config"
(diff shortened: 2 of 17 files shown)


server/server.go
```diff
@@ -0,0 +1,100 @@
...
+		}
+	}
+}
+
+func proxyRequest(client *api.Client, r *http.Request, body []byte) (*api.Response, error) {
+	// http.Transport will transparently request gzip and decompress the response, but only if
+	// the client doesn't manually set the header. Removing any Accept-Encoding header allows the
+	// transparent compression to occur.
+	r.Header.Del("Accept-Encoding")
+	client.SetHeaders(r.Header)
```
ncabatoff, line 70: I don't think you need to call SetHeaders here, though if you remove this we should sort out whether X-Vault-Request should be populated by your proxy.  Actually we need to work that out regardless.

server/server.go
```diff
@@ -0,0 +1,100 @@
...
+				return
+			}
+
+			err = resp.Write(w)
+			if err != nil {
+				http.Error(w, "failed to write response back to requester", 500)
+				return
+			}
+			logger.Printf("Successfully proxied %s %s\n", r.Method, r.URL.Path)
+		default:
```
ncabatoff, line 58: Out of curiosity, why only GETs?

main.go
```diff
@@ -3,117 +3,185 @@ package main
...
 
-	if vaultAddr == "" || vaultAuthProvider == "" || vaultAuthRole == "" || len(configuredSecrets) == 0 {
-		logger.Fatal("missing VAULT_ADDR, VAULT_AUTH_PROVIDER, VAULT_AUTH_ROLE, or VAULT_SECRET_ environment variables.")
+	err = writePreconfiguredSecrets(client)
+	if err != nil {
+		return nil, err
 	}
 
-	client, err := vault.NewClient(logger, vaultAuthRole, vaultAuthProvider, vaultIamServerId)
+	srv := server.New(client)
```
catsby, line 98: I think we're missing some inputs here? 

```
./main.go:98:19: not enough arguments in call to server.New
        have (*api.Client)
        want (*log.Logger, *api.Client)
```

main.go
```diff
@@ -3,117 +3,185 @@ package main
...
+	return srv, nil
+}
+
+func writePreconfiguredSecrets(client *api.Client) error {
+	configuredSecrets, err := config.ParseConfiguredSecrets()
 	if err != nil {
-		logger.Fatalf("error getting client: %s", err)
-	} else if client == nil {
-		logger.Fatalf("nil client returned: %s", err)
+		return fmt.Errorf("Failed to parse configured secrets to read: %s", err)
```
catsby, line 137: nit: errors should not be capitalized. This happens in a few other places

main.go
```diff
@@ -3,117 +3,185 @@ package main
...
 
-	configuredSecrets, err := config.ParseConfiguredSecrets()
+	if vaultAddr == "" || vaultAuthProvider == "" || vaultAuthRole == "" {
+		return nil, errors.New("missing VAULT_ADDR, VAULT_AUTH_PROVIDER or VAULT_AUTH_ROLE environment variables")
+	}
+
+	client, err := vault.NewClient(logger, vaultAuthRole, vaultAuthProvider, vaultIAMServerID)
 	if err != nil {
-		logger.Fatalf("Failed to parse configured secrets to read: %s", err)
+		return nil, fmt.Errorf("error getting client: %s", err)
```
catsby, line 88: We should use error wrapping here and other `fmt.Errorf` calls where we give an error
```suggestion
		return nil, fmt.Errorf("error getting client: %w", err)
```

server/server.go
```diff
@@ -0,0 +1,100 @@
...
+			body, err := ioutil.ReadAll(r.Body)
+			if err != nil {
+				http.Error(w, err.Error(), 500)
+				return
+			}
+
+			logger.Printf("Proxying %s %s\n", r.Method, r.URL.Path)
+			resp, err := proxyRequest(cl, r, body)
+			if err != nil {
+				http.Error(w, err.Error(), 502)
```
catsby, line 48:
```suggestion
				http.Error(w, err.Error(), http.StatusBadGateway)
```

catsby (commented):
Minor feedback for now, all cosmetic though! I'll take another pass tomorrow when I have additional time to run the tests

server/server.go
```diff
@@ -0,0 +1,100 @@
...
+	return &srv
+}
+
+// The proxyHandler is based on the Send function from Vault Agent's proxy:
+// https://github.com/hashicorp/vault/blob/22b486b651b8956d32fb24e77cef4050df7094b6/command/agent/cache/api_proxy.go
+func proxyHandler(logger *log.Logger, vaultClient *api.Client) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
+		case "GET":
+			cl, err := vaultClient.Clone()
```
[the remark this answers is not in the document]
tomhjp: Thanks, yeah I had assumed the `client.SetHeaders` was another reason for the clone but I just checked the implementation of that and you're right.

server/server.go
tomhjp: We were taking a cautious approach to what we would support, but tbh it is a very artificial limitation, so I've removed it.

server/server.go
tomhjp: One use case I had in mind was setting namespace in the request headers, and I assumed there would be others I didn't know about so thought it would be best to just honour the original HTTP request as closely as possible. Do you think we should handle cases like namespace explicitly?

Good question about the `X-Vault-Request` header, I need to learn a bit more about the design/intention of that before I can answer your question.

server/server.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+
+			err = resp.Write(w)
+			if err != nil {
+				http.Error(w, "failed to write response back to requester", 500)
+				return
+			}
+			logger.Printf("Successfully proxied %s %s\n", r.Method, r.URL.Path)
+		default:
+			http.Error(w, fmt.Sprintf("unsupported method for vault-lambda-extension: %s", r.Method), 400)
```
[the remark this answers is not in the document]
tomhjp: Yeah perhaps we should just remove this artificial limitation - it's extra code for no real benefit.

server/server.go
```diff
@@ -0,0 +1,100 @@
...
+	fwReq.BodyBytes = body
+
+	query := r.URL.Query()
+	if len(query) > 0 {
+		fwReq.Params = query
+	}
+
+	resp, err := client.RawRequestWithContext(r.Context(), fwReq)
+	if err != nil {
+		errString := fmt.Sprintf("Failed to proxy request: %s", err)
```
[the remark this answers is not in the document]
tomhjp: This is definitely the area that I felt was weakest in the PR, so I welcome any comments on it. Handling errors is a little tricky because we are acting as an HTTP-level proxy, but the error-handling behaviour of the Vault API client is significantly different from that of the underlying Golang HTTP client. I've reworked the error handling in an attempt to improve the behaviour when Vault returns non-OK error codes, but it's quite confusing and error-prone trying to unwind the additional error handling that `api` does.

server/server.go
```diff
@@ -0,0 +1,100 @@
...
+		if resp != nil {
+			defer resp.Body.Close()
+			body, err := ioutil.ReadAll(resp.Body)
+			errString += fmt.Sprintf("\n\n(%d) %s", resp.StatusCode, string(body))
```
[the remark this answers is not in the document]
tomhjp: hah thank you, yeah I knew this wasn't right but failed to improve it on my own. I think you're right that it should be using `multierror` or error wrapping instead, but as it turns out I just deleted all of this special handling in response to the above comment anyway. The errors we're handling here include things like 403 from Vault, so we actually want to pass those responses on without modification.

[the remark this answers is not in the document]
tomhjp: On the format of the body content, we don't concern ourselves with that, we just read bytes from the request body, forward them on as bytes, and do the same on the way back. I think it will almost always happen to be JSON? But we copy headers from the request and the response too, so that's all handled outside our proxy server.

main.go
tomhjp: Oops not sure how that happened. Should be fixed in the latest commits.

server/server.go
ncabatoff: Oh I agree that we should honour the incoming request's headers, I'm just saying you don't need to call client.SetHeaders for that to happen.  You can simply set fwReq.Headers = r.Header.  I don't know about namespaces, but if you don't have a specific requirement regarding them I probably wouldn't pay them any mind, just pass the requests through largely unchanged and everything should be fine.

main.go
```diff
@@ -3,117 +3,169 @@ package main
...
 func main() {
 	logger := log.New(os.Stderr, fmt.Sprintf("[%s] ", extensionName), log.Ldate|log.Ltime|log.LUTC)
 	ctx, cancel := context.WithCancel(context.Background())
+	extensionClient := extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
+	_, err := extensionClient.Register(ctx, extensionName)
+	if err != nil {
+		logger.Fatal(err)
+	}
+
+	threadFinishedCh := make(chan struct{}, 2)
```
catsby, line 37: it may be worth documenting why this is a buffered channel of size 2, it's not immediately obvious to me and the double `<-threadFinishedCh` lower in this function isn't clear

main.go
```diff
@@ -3,117 +3,169 @@ package main
...
-	processEvents(ctx, logger)
+	// Ensure we wait for the HTTP server to gracefully shut down.
+	// One signal from the HTTP server thread and one from the shutdown thread.
+	<-threadFinishedCh
+	<-threadFinishedCh
+	logger.Println("Graceful shutdown complete")
 }
 
-func initialiseExtension(logger *log.Logger) {
+func initialiseExtension(logger *log.Logger, threadFinishedCh chan struct{}) (*http.Server, error) {
```
catsby, line 74: It may be appropriate to rename this to `RunExtension` or something, as the purpose changes from "writing the secrets to disk and quitting" to "run the proxy until things error or we shut down"

main.go
```diff
@@ -3,117 +3,169 @@ package main
...
 	extensionName = "vault-lambda-extension"
 )
 
-var (
-	extensionClient = extension.NewClient(os.Getenv("AWS_LAMBDA_RUNTIME_API"))
-)
-
 func main() {
 	logger := log.New(os.Stderr, fmt.Sprintf("[%s] ", extensionName), log.Ldate|log.Ltime|log.LUTC)
 	ctx, cancel := context.WithCancel(context.Background())
```
catsby, line 30: We use this `ctx` to cleanly shut down our `processEvents` function, should it also be passed into the modified `initialiseExtension` to cleanly shut down the server?

catsby (commented):
Added a few more comments

server/server.go
tomhjp: Ahh that makes sense, yeah that's better, thanks.

server/server.go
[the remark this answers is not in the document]
tomhjp: Based on my comment above and some offline discussion, I've now reworked the proxying to be at the HTTP level instead of using the vault client. I think the error handling is a lot more robust and easier to inspect now. LMK what you think.

README.md
```diff
@@ -100,15 +100,13 @@ Environment variable    | Description | Required | Example value
...
+`VAULT_CLIENT_TIMEOUT`  | Timeout for Vault requests. Default value is 60s. Ignored by proxy server. **Any value over 10s will exceed the Extensions API timeout and therefore have no effect** | No | `5s`
 `VAULT_IAM_SERVER_ID`   | Value to pass to the Vault server via the [`X-Vault-AWS-IAM-Server-ID` HTTP Header for AWS Authentication](https://www.vaultproject.io/api-docs/auth/aws#iam_server_id_header_value) | No | `vault.example.com`
-`VAULT_MAX_RETRIES`     | Maximum number of retries on `5xx` error codes. Defaults to 2 | No | `2`
+`VAULT_MAX_RETRIES`     | Maximum number of retries on `5xx` error codes. Defaults to 2. Ignored by proxy server | No | `2`
 `VAULT_SKIP_VERIFY`     | Do not verify Vault's presented certificate before communicating with it. Setting this variable is not recommended and voids Vault's [security model][vault-security-model]  | No | `true`
 `VAULT_TLS_SERVER_NAME` | Name to use as the SNI host when connecting via TLS | No | `vault.example.com`
-`VAULT_RATE_LIMIT`      | Only applies to a single invocation of the extension. See [Vault Commands (CLI)][vault-env-vars] documentation for details | No | `10`
```
tomhjp, line 108: I've undocumented `VAULT_RATE_LIMIT`, `VAULT_SRV_LOOKUP` and `VAULT_MFA`. The first because it's basically irrelevant for this use case and just adds noise to the documentation. The latter two because it turns out we actually don't support them right now and this PR doesn't add support.

main.go
ncabatoff: It might also be worth considering using a sync.WaitGroup instead of a channel for this.   It tends to make it easier to manage error cases, since the place where you signal completion (.Done) is closer to where you actually start a goroutine and call .Add.

tomhjp: Good idea, done 👍

main.go
tomhjp: I'm not aware of a way to use a context to directly cause an HTTP server to shutdown. However, we do handle clean shutdown in the goroutine below by calling `srv.Shutdown` when we receive an interrupt signal or a SHUTDOWN event from the runtime API.

Your comment got me looking closely though and one pre-existing place we don't handle shutdown well is we don't pass a context to the Vault login or read calls for `writePreconfiguredSecrets`. However, I'm not sure how to handle this. We can't start listening for SHUTDOWN from the runtime API, because to register for those events would signal our readiness too early. So all we could handle is receiving OS shutdown signals, but I'm not sure in what scenario we would get those? Perhaps if the runtime API wants to shutdown the extension before it has initialised? I would guess that could happen if the extension takes longer than 10s to startup.

Either way though, this PR doesn't make that bit worse, so I'll probably save those changes for another PR if that all sounds right to you.

server/server.go
```diff
@@ -0,0 +1,81 @@
...
+func proxyHandler(logger *log.Logger, config *api.Config, token string) func(http.ResponseWriter, *http.Request) {
+	return func(w http.ResponseWriter, r *http.Request) {
+		logger.Printf("Proxying %s %s\n", r.Method, r.URL.Path)
+		fwReq, err := proxyRequest(r, config.Address, token)
+		if err != nil {
+			http.Error(w, fmt.Sprintf("failed to generate proxy request: %s", err), http.StatusInternalServerError)
+			return
+		}
+
+		resp, err := config.HttpClient.Do(fwReq)
```
ncabatoff, line 36: Any reason not to use retryablehttp.Client here?

tomhjp: I don't have strong opinions here, but I erred on the side of simplicity in order to delegate more complex behaviour like retrying and error handling to the proxy's client. The less the proxy does, the more transparent it is to the client who can mostly act as though it's talking directly to Vault. Do you have a preference to add retries?

tomhjp: After syncing offline, we decided to leave this as-is for the first version. We may add retries in a future revision, as long as we can be confident about what requests are safe to retry (idempotent etc).

tomhjp:
I've just updated the demo function and run this through the quick-start demo:

```text
START RequestId: 53da3758-dbc0-4ae4-a9a7-7ce820f45762 Version: $LATEST
[vault-lambda-extension] 2021/02/26 17:52:39 Received event
[vault-lambda-extension] 2021/02/26 17:52:39 Waiting for event...
[runtime] Received invocation: {}
[runtime] Executing function: main
[demo-function] Received: 
[demo-function] Reading file /tmp/vault_secret.json
[demo-function] Querying users using credentials from disk
[demo-function] users: 
[demo-function]      vaultadmin
[demo-function]      rdsadmin
[demo-function]      v-aws-vaul-lambda-f-OWXSuL4C1NaVKgjTiX1E-1614361359
[demo-function]      v-aws-vaul-lambda-f-95CvfsYkfe6Akrs3GtlZ-1614361959
[vault-lambda-extension] 2021/02/26 17:52:41 Proxying GET /v1/database/creds/lambda-function
[vault-lambda-extension] 2021/02/26 17:52:41 Successfully proxied GET /v1/database/creds/lambda-function
[demo-function] Querying users using credentials from proxy
[demo-function] users: 
[demo-function]      vaultadmin
[demo-function]      rdsadmin
[demo-function]      v-aws-vaul-lambda-f-OWXSuL4C1NaVKgjTiX1E-1614361359
[demo-function]      v-aws-vaul-lambda-f-95CvfsYkfe6Akrs3GtlZ-1614361959
[demo-function]      v-aws-vaul-lambda-f-8fXNZEDXPVBe6USQYGWf-1614361961
END RequestId: 53da3758-dbc0-4ae4-a9a7-7ce820f45762
REPORT RequestId: 53da3758-dbc0-4ae4-a9a7-7ce820f45762  Duration: 1286.11 ms    Billed Duration: 1415 ms        Memory Size: 128 MB     Max Memory Used: 62 MB  Init Duration: 127.97 ms
```

README.md
```diff
@@ -100,15 +110,12 @@ Environment variable    | Description | Required | Example value
 `VAULT_CAPATH`          | Path to a _directory_ of PEM-encoded CA certificate files on the local disk | No | `/tmp/certs`
 `VAULT_CLIENT_CERT`     | Path to a PEM-encoded client certificate on the local disk | No | `/tmp/client.crt`
 `VAULT_CLIENT_KEY`      | Path to an unencrypted, PEM-encoded private key on disk which corresponds to the matching client certificate | No | `/tmp/client.key`
-`VAULT_CLIENT_TIMEOUT`  | Timeout for Vault requests. Default value is 60s. **Any value over 10s will exceed the Extensions API timeout and therefore have no effect** | No | `5s`
-`VAULT_IAM_SERVER_ID`   | Value to pass to the Vault server via the [`X-Vault-AWS-IAM-Server-ID` HTTP Header for AWS Authentication](https://www.vaultproject.io/api-docs/auth/aws#iam_server_id_header_value) | No | `vault.example.com`
-`VAULT_MAX_RETRIES`     | Maximum number of retries on `5xx` error codes. Defaults to 2 | No | `2`
+`VAULT_CLIENT_TIMEOUT`  | Timeout for Vault requests. Default value is 60s. Ignored by proxy server. **Any value over 10s will exceed the Extensions API timeout and therefore have no effect** | No | `5s`
```
kalafut, line 113: Is default 60s or 5s?

main.go
```diff
@@ -3,117 +3,157 @@ package main
...
-		logger.Fatalf("nil client returned: %s", err)
+		return nil, err
 	}
 
-	if _, err = os.Stat(config.DefaultSecretDirectory); os.IsNotExist(err) {
-		err = os.MkdirAll(config.DefaultSecretDirectory, 0755)
-		if err != nil {
-			logger.Fatalf("Failed to create directory %s: %s", config.DefaultSecretDirectory, err)
-		}
+	ln, err := net.Listen("tcp", "127.0.0.1:8200")
```
kalafut, line 98: Can this port ever conflict with anything (i.e. need to be configurable?)

README.md
tomhjp: Yeah that's right. It's just the default from the `api` package's default client and a tricky one to handle in this context. I don't think we can really manage a better default because if we set it to 10s then it still has no effect because AWS' 10s timer will start before ours. And if we set it below 10s then we are just guessing at how much time we have left and are arbitrarily introducing failures.

main.go
tomhjp: Given the context of Lambdas, I figured it was highly unlikely to ever cause conflicts, but it is an easy one to add configuration for if we get that request. Do you think it would be better to add that configuration upfront?

tomhjp: (The implied answer to your question is "yes it's possible" 🙂)

main.go
catsby: I'm not sure if the code changed or I just didn't read it well enough, I think my concern of the server _not_ being shut down if the interrupt or shutdown were received is at the least no longer valid, I see the `cancel()` and then `srv.Shutdown` now. I was thinking that `initialiseExtension` needed to listen for `ctx.Done` but I see now we should be good with regards to the shutdown

catsby (approved):

main.go
```diff
@@ -3,117 +3,157 @@ package main
...
+	if err != nil {
+		logger.Fatal(err)
+	}
 
-	sigs := make(chan os.Signal, 1)
-	signal.Notify(sigs, syscall.SIGTERM, syscall.SIGINT)
+	shutdownChannel := make(chan struct{})
 	go func() {
-		s := <-sigs
+		wg.Add(1)
```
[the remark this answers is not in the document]
tomhjp: Oops, good catch, thanks.

main.go
kalafut: We can defer.

PR #28, opened 2021-03-01
Auto-renew Vault token for server mode

tomhjp:
Checks the Vault client's auth token and renews/re-authenticates to Vault when it gets near expiry.

Follows up on #27 to finish the work to address #13

Acceptance/integration testing done:
- [x] Add an integration test using docker-compose
- [x] Some manual acceptance testing with real infrastructure and the quick-start guide


vault/client.go
```diff
@@ -27,58 +31,177 @@ func NewClient(logger *log.Logger, vaultAuthRole, vaultAuthProvider string, vaul
...
+	for {
+		// Sleep for the first 2/3 of the initial token TTL, as the renewer below
+		// will immediately start with a renew call, and in many cases we expect
+		// the Lambda function will exit before the first renewal is even needed.
+		err := sleepUntilRenewingShouldStart(ctx, secret)
+		if err != nil {
+			logger.Printf("error while waiting for first token renewal: %s\n", err)
+		}
+
+		err = renewToken(ctx, logger, client, secret)
```
catsby, line 107: Does this function block? I think it does, it might be good to document that at the function definition and maybe even call that out here

vault/client.go
```diff
@@ -27,58 +31,177 @@ func NewClient(logger *log.Logger, vaultAuthRole, vaultAuthProvider string, vaul
 	ses := session.Must(session.NewSession())
 	stsSvc := sts.New(ses)
 
+	logger.Println("attempting Vault login")
+	secret, err := login(ctx, stsSvc, vaultClient, authConfig)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Start background threads to renew and re-auth as required.
```
catsby, line 40: nit:

```suggestion
	// Start background goroutine to renew and re-auth as required.
```

vault/client_test.go
```diff
@@ -0,0 +1,206 @@
...
+			r := <-vaultCh
+			vaultCalls = append(vaultCalls, r)
+			r = <-vaultCh
+			vaultCalls = append(vaultCalls, r)
+			cancel()
+		}()
+		require.NoError(t, renewToken(ctx, logger, client, with1sLease))
+		require.Contains(t, vaultCalls[0].URL.Path, "auth/token/renew-self")
+		require.Contains(t, vaultCalls[1].URL.Path, "auth/token/renew-self")
+		require.Contains(t, string(logBuffer.Bytes()), "successfully renewed token")
```
catsby, line 84:
```suggestion
		require.Contains(t, logBuffer.String(), "successfully renewed token")
```

catsby (approved):
👍 minor nit picks but otherwise looked solid on first pass.

vault/client.go
```diff
@@ -1,84 +1,167 @@
...
+	VaultConfig *api.Config
+
+	logger     *log.Logger
+	stsSvc     *sts.STS
+	authConfig config.AuthConfig
+
+	// Token refresh/renew data.
+	tokenExpiry    time.Time
+	tokenTTL       time.Duration
+	tokenRenewable bool
```
jasonodonnell, line 33: nit: I wonder if it would be cleaner to just store the `*api.Secret` representing the token. All of these fields should be there, I think.

vault/client.go
```diff
@@ -1,84 +1,167 @@
...
+
+// Returns true if current time is after tokenExpiry, or within 10s.
+func (c *Client) expired() bool {
+	return time.Now().Add(10 * time.Second).After(c.tokenExpiry)
+}
+
+// Returns true if tokenExpiry time is in less than 20% of tokenTTL.
+func (c *Client) shouldRenew() bool {
+	remaining := c.tokenExpiry.Sub(time.Now())
+	return c.tokenRenewable && remaining.Nanoseconds() < c.tokenTTL.Nanoseconds()/5
```
jasonodonnell, line 166: I'm wondering if doing a token lookup self and getting the actual time left instead of storing metadata is a more reliable solution. Tracking time in the function could be subject to clock skew, especially when TTLs are small. One downside to lookups is we increase the traffic to Vault, but it's likely more reliable in the end.

tomhjp: This is a good thought, and definitely worth considering. Of course the same comment also applies to the logic in `expired()` too.

On clock skew, I don't think we need to worry about cross-server clock skew because this value is calculated from a TTL duration and then only consumed within the same execution environment. Obviously that's not the same as a VM, but hopefully an execution environment's clock is more reliable at tracking a duration than the clock skew that can happen between servers.

Given that, the best we can do is eliminate the clock drift that could happen between 0 and 80% of the TTL. I don't have data, but I would hope that's normally insignificant compared to the drift due to network latency between Lambda and Vault. That network latency error could be large as we calculate the expiry time from the time we receive the response, instead of the unknown time Vault sent it. So eliminating clock drift is probably not enough gain to justify the token-lookup latency on every request, which is pure compute cost in a Lambda function.

On the topic of small TTLs, we will only ever attempt to renew if the token is not considered "expired", which means it must have at least 10s of TTL left. I think that should serve most deployments well, but thinking this through has convinced me it would be good to make that 10s grace window configurable.

vault/client.go
tomhjp: Surprisingly, getting the token TTL from an `*api.Secret` can return an error, so extracting it early helps us to fail nice and early if there will be a problem. Also, even if we stored the `*api.Secret`, we would still have to have a separate field to mark the time it was fetched, because the TTL is given as a duration, not a time. Overall I think this is a bit "six of one, half a dozen of the other".

vault/client.go
```diff
@@ -1,84 +1,167 @@
...
-		return nil, nil, fmt.Errorf("got no response from the %s authentication provider", vaultAuthProvider)
+
+	token, err := secret.TokenID()
+	if err != nil {
+		return fmt.Errorf("error reading token: %s", err)
+	}
+	c.VaultClient.SetToken(token)
+
+	err = c.updateTokenMetadata(secret)
+	if err != nil {
```
catsby, line 128: nit: could just do `return c.updateTokenMetadata(secret)`

vault/client.go
```diff
@@ -1,84 +1,167 @@
...
+		authConfig: authConfig,
+	}
+
+	return client, nil
+}
+
+// Token synchronously renews/re-auths as required and returns a Vault token.
+func (c *Client) Token(ctx context.Context) (string, error) {
+	c.mtx.Lock()
+	defer c.mtx.Unlock()
```
catsby, line 62: Minor nit: every request handled by the `proxyHandler` is going to capture a full lock here. It might not be a problem though. If it turns out to be a problem we'd likely want to do use a read/write mutex and "upgrade" the lock here (it's not really upgrading but you get my point). 

No action required here, just posting

catsby (commented):
Minor nits but otherwise looks great! I did not explicitly "approve" because I haven't been able to take the time to run the tests or try it out myself yet, but I wanted to possibly unblock you if others have checked out the tests

vault/client.go
catsby: After more reflection, I'm not sure this is really something worth looking into at all

tomhjp: I went through the same thought process as you 🙂 I actually started writing this with a `RWMutex` and then decided it was probably too early to make that optimisation without some data. Assuming a reasonable token TTL and a scenario where traffic is high enough that we care about lock contention, most calls to this function will already have a valid token that doesn't need renewing, so I think it will probably be quick enough.

vault/client.go
tomhjp: I've now added a configurable for token TTL grace period.

jasonodonnell (approved):

Issue #18, opened 2020-11-25
Unable to access Vault secrets due to incorrect AWS region


dipack95:
I'm trying to use this Lambda extension to fetch secrets from Vault, before the code in Lambda function runs. This is the error I'm receiving:

```text
START RequestId: 21b2eabf-0f3e-4ff2-8dca-932a6041e245 Version: $LATEST
8cd0a2c4-942c-4cbc-b6a2-0de422a07ba6[vault-lambda-extension] 2020/11/25 23:05:44 Initialising
[vault-lambda-extension] 2020/11/25 23:05:44 attemping Vault login...
[vault-lambda-extension] 2020/11/25 23:05:45 error getting client: Error making API request.

URL: PUT https://vault.xx.com/v1/auth/aws/login
Code: 400. Errors:

* error making upstream request: received error code 403 from STS: <ErrorResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
  <Error>
    <Type>Sender</Type>
    <Code>SignatureDoesNotMatch</Code>
    <Message>Credential should be scoped to a valid region, not 'us-east-1'. </Message>
  </Error>
  <RequestId>7df68af3-8edc-44a3-b75c-230b1b36dccc</RequestId>
</ErrorResponse>
EXTENSION	Name: vault-lambda-extension	State: Registered	Events: [INVOKE,SHUTDOWN]
END RequestId: 21b2eabf-0f3e-4ff2-8dca-932a6041e245
REPORT RequestId: 21b2eabf-0f3e-4ff2-8dca-932a6041e245	Duration: 5002.45 ms	Billed Duration: 5100 ms	Memory Size: 128 MB	Max Memory Used: 37 MB	
XRAY TraceId: 1-5fbee346-546a8a1d79b0e587206cea7f	SegmentId: 75925ef155507a5a	Sampled: true	
RequestId: 21b2eabf-0f3e-4ff2-8dca-932a6041e245 Error: exit status 1
Extension.Crash
```

This is reproducible on my local machine as well:

```bash
~/projects  aws-vault exec identity -- vault login --method=aws role=developer_role                                                                                                                                 
Error authenticating: Error making API request.

URL: PUT https://vault.xx.com/v1/auth/aws/login
Code: 400. Errors:

* error making upstream request: received error code 403 from STS: <ErrorResponse xmlns="https://sts.amazonaws.com/doc/2011-06-15/">
  <Error>
    <Type>Sender</Type>
[...]

PR #29, opened 2021-03-19
Update docs for server mode, notices and regional STS

tomhjp:
A collection of documentation updates:

* Adds missing public preview label and responsible disclosure notice
* Addresses the issue in #18 
* Updates limitations notice and recommended way to use the extension

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,11 +29,12 @@ The extension authenticates with Vault using [AWS IAM auth][vault-aws-iam-auth],
 and all configuration is supplied via environment variables. There are two methods
 to read secrets, which can both be used side-by-side:
 
+* **Recommended**: Make unauthenticated requests to the extension's local proxy
+  server at `http://127.0.0.1:8200`, which will add an authentication header and
+  proxy to the configured `VAULT_ADDR`. Responses from Vault are returned without
+  modification.
 * Configure environment variables such as `VAULT_SECRET_PATH` for the extension
   to read a secret and write it to disk.
-* **Recommended**: Make unauthenticated requests to the extension's local server at
-  `http://127.0.0.1:8200`, which will add an authentication header and proxy to
-  the configured `VAULT_ADDR`.
 
 ## Getting Started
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -20,7 +31,7 @@ to read secrets, which can both be used side-by-side:
 
 * Configure environment variables such as `VAULT_SECRET_PATH` for the extension
   to read a secret and write it to disk.
-* Make unauthenticated requests to the extension's local server at
+* **Recommended**: Make unauthenticated requests to the extension's local server at
```
catsby, line 34: If we do recommend this, let's list it first

catsby: We may also want to mention that with this method you're getting a response just like you would if you were talking to Vault directly, where as the `VAULT_SECRET_PATH` approach dumps values to files. With the Server method they're getting `api.Secret` values in an HTTP response

README.md
```diff
@@ -118,19 +131,44 @@ Environment variable    | Description | Required | Example value
...
+Secrets written to disk or returned from the proxy server will not be automatically
+refreshed when they expire. This is particularly important if you configure the
+extension to write secrets to disk, because the extension will only write to disk
+once per execution environment, rather than once per function invocation. If you
+use [provisioned concurrency][lambda-provisioned-concurrency] or if your Lambda
+is invoked often enough that execution contexts live beyond the lifetime of the
+secret, then secrets on disk are likely to become invalid. In line with
+[Lambda best practices][lambda-best-practices], we recommend avoiding writing
+secrets to disk where possible, and exclusively consuming secrets via the proxy
+server.
```
catsby, line 162: I think it's implied that the proxy server method will refresh secrets here? This paragraph doesn't directly say that though, and mentions secrets returned from proxy method, so it's confusing to me if those secrets are refreshed or not? 

Are they refreshed when requested again?

catsby (commented):
Looks good overall, but I think we should add some clarity about the refreshing

README.md
tomhjp: Good points, we actually write an `api.Secret` out to disk for the pre-configured approach, so that data returned is very similar, just consumed by the lambda function differently.

README.md
tomhjp: Yeah I struggled to explain this clearly. I've given it another pass, LMK what you think.

catsby (approved):
👍 LGTM

PR #30, opened 2021-04-16
Allow customizing the sts region via an env var

dmattia:
Hello again @tomhjp 👋 

As an update on https://github.com/hashicorp/vault-lambda-extension/issues/18, we have been using `AWS_STS_REGIONAL_ENDPOINTS=regional` as you suggested for a while now with great success 😄 

Unfortunately, some of our lambda services are now becoming multi-regional, but we want them to authenticate to the vault role, which always requires us to use the same sts endpoint.

As a workaround, we could create multiple roles in different regions, but that is a lot of duplication when we want all regions to be able to access the same exact secrets in vault.

Would you be willing to use the original approach you described in this commend: https://github.com/hashicorp/vault-lambda-extension/issues/18#issuecomment-787972647 where you suggested having an env var that would allow for overriding the sts endpoint when specified?

I did my best to implement that idea in this PR

Thank you for all your help, as always!


vault/client.go
```diff
@@ -19,6 +21,7 @@ import (
 
 const (
 	tokenExpiryGracePeriodEnv     = "VAULT_TOKEN_EXPIRY_GRACE_PERIOD"
+	stsEndpointRegionEnv          = "VAULT_STS_ENDPOINT_REGION_ENV"
```
tomhjp, line 24:
```suggestion
	stsEndpointRegionEnv          = "VAULT_STS_ENDPOINT_REGION"
```

tomhjp (commented):
Thanks for raising this. IIUC this is essentially needed because e.g. you want a Lambda running in `us-east-1` to be able to use a Vault AWS IAM auth mount that is configured to talk to `us-west-1`, but the `AWS_REGION` variable that controls which STS endpoint the extension talks to is immutable?

Seems like a reasonable ask and mostly LGTM. Just one nit, and it would be good to have an entry in the readme table for this configurable please! :)

dmattia:
@tomhjp Thanks for the review! I updated the env name and added some docs to the Readme as you suggested

README.md
```diff
@@ -113,6 +113,7 @@ Environment variable              | Description | Required | Example value
 `VAULT_SECRET_PATH_FOO`           | Additional secret path to read, where FOO can be any name, as long as a matching `VAULT_SECRET_FILE_FOO` is specified | No | `secret/lambda-app/token`
 `VAULT_SECRET_FILE_FOO`           | Must exist for any correspondingly named `VAULT_SECRET_PATH_FOO`. Name has no further effect beyond matching to the correct path variable | No | `/tmp/token`
 `VAULT_TOKEN_EXPIRY_GRACE_PERIOD` | Period at the end of the proxy server's auth token TTL where it will consider the token expired and attempt to re-authenticate to Vault. Must have a unit and be parseable by `time.Duration`. Defaults to 10s. | No | `1m`
+`VAULT_STS_ENDPOINT_REGION`       | The region of the STS regional endpoint to authenticate with. If the AWS IAM auth mount uses a regional STS endpoint, then this needs to match the region of the endpoint. Defaults to using the global endpoint, or the region the lambda resides in if `AWS_STS_REGIONAL_ENDPOINTS` is set to `regional` | No | `eu-west-1`
```
tomhjp, line 116:
```suggestion
`VAULT_STS_ENDPOINT_REGION`       | The region of STS to authenticate with. If the AWS IAM auth mount specified uses a regional STS endpoint, then this needs to match the region of that endpoint. Defaults to using the global endpoint, or the region the Lambda resides in if `AWS_STS_REGIONAL_ENDPOINTS` is set to `regional` | No | `eu-west-1`
```

tomhjp:
Thanks for addressing those comments. I've built and tested this manually, and confirmed it works as expected. That includes `AWS_STS_REGIONAL_ENDPOINTS` being unnecessary/having no effect when the new env var is in place.

One last thing I forgot in my previous review was please could you also add a changelog entry? As a suggestion, my best shot at summarising this succinctly:

```
Add `VAULT_STS_ENDPOINT_REGION` to make STS regional endpoint used for auth configurable separately from the region Lambda is deployed in. [[GH-30](https://github.com/hashicorp/vault-lambda-extension/pull/30)]
```

You can create a new features heading similar to the existing example. Otherwise LGTM!

dmattia:
Awesome, thanks for testing @tomhjp!

Just as a note, I will be out of office for a little over a week starting tomorrow. Feel free to merge in the meantime and make any changes you'd like, otherwise I will be back to finish this off in May 😄

PR #34, opened 2021-05-05
Add instructions for deploying in container image

tomhjp:
Give instructions on how to deploy the extension into AWS Lambda functions that are package as container images.

I'm open to feedback on how to better set this up for `quick-start` in particular. The changes required to the guide are non-trivial, but not so great that I felt it was necessary to take a copy of the whole guide and fork it, which will cause maintenance problems.


README.md
```diff
@@ -38,6 +38,11 @@ to read secrets, which can both be used side-by-side:
 
 ## Getting Started
 
+The [learn guide][vault-learn-guide] is the most complete and fully explained
```
jasonodonnell, line 41: This looks like an error. Should it be `[learn guide](vault-learn-guide)`?

quick-start/demo-function/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Based on vanilla Dockerfile to build a Golang image from:
+# https://docs.aws.amazon.com/lambda/latest/dg/go-image.html
+# The only significant modification is to copy the extension binary into the image.
+FROM public.ecr.aws/lambda/provided:al2 as build
```
jasonodonnell, line 4: Is there a hard requirement to use this Docker image for builds? This Dockerfile uses a builder pattern to produce a binary, so does it make more sense to use the Golang Docker image instead?

I thing the Golang docker image might be a bit better for maintenance because right now `yum` is just installing latest version of Golang, but the Docker image would be versioned with a specific Golang installation.

README.md
tomhjp: Not an error, I just double checked and the link does work. This is an alternative way to specify links in markdown. `vault-learn-guide` references a label for a URL at the bottom of this document.

quick-start/demo-function/README.md
```diff
@@ -0,0 +1,43 @@
...
+  ```bash
+  pushd ../../ && GOOS=linux GOARCH=amd64 go build -ldflags '-s -w' -a -o quick-start/demo-function/pkg/extensions/vault-lambda-extension main.go && popd
+  docker build -t demo-function .
+  ```
+
+* Tag the image, and push it to your new ECR repository:
+
+  ```bash
+  docker tag demo-function:latest <account>.dkr.ecr.<region>.amazonaws.com/demo-function:latest
+  docker push <account>.dkr.ecr.<region>.amazonaws.com/demo-function:latest
```
jasonodonnell, line 37: Nit, don't feel obligated to adopt, but this is a little bit nicer for the user I think:
```suggestion
  export AWS_ACCOUNT="ACCOUNT HERE"
  export AWS_REGION="REGION HERE"
  docker tag demo-function:latest ${AWS_ACCOUNT?}.dkr.ecr.${AWS_REGION?}.amazonaws.com/demo-function:latest
  docker push ${AWS_ACCOUNT?}.dkr.ecr.${AWS_REGION?}.amazonaws.com/demo-function:latest
```

quick-start/demo-function/Dockerfile
tomhjp: Yeah I think that's a worthwhile improvement. As per the comment, this file is almost entirely just lifted from AWS docs because the only thing it's really trying to demonstrate is that the only additional step you need for extensions is the copy at the bottom (the rest is setting up the function packaging), but it does make more sense to use a versioned image for the builder.

quick-start/demo-function/README.md
```diff
@@ -0,0 +1,43 @@
...
+    name                 = "demo-function"
+  }
+  ```
+
+* Package the extension and function together into a Docker image. From this
+  directory:
+
+  ```bash
+  pushd ../../ && GOOS=linux GOARCH=amd64 go build -ldflags '-s -w' -a -o quick-start/demo-function/pkg/extensions/vault-lambda-extension main.go && popd
+  docker build -t demo-function .
```
jasonodonnell, line 30: Wondering if we should just have a Makefile and do something like `make build`?

README.md
jasonodonnell: Cool, thanks!

quick-start/demo-function/Dockerfile
jasonodonnell: It's up to you, but maybe we should be more explicit with the version of Golang on `yum install` if we're going to keep this image.

jasonodonnell (approved):
Some minor things but overall I think the direction of this PR is fine.

README.md
tomhjp: BTW the reason I am using it is mainly as a nice way reference the same link multiple times, but I also like how it means all the links can get grouped together and they keep the noise out of the main text of the markdown source.

quick-start/demo-function/README.md
tomhjp: I didn't want to screw around with the well-tested `build.sh` script too much, and would have wanted to fold it into the Makefile, so instead I added a `build-container.sh` script alongside `build.sh`. That keeps the instructions much simpler.

PR #42, opened 2021-07-01
Update released version to tag v0.4.0, layer version 10

tomhjp:
New version released earlier today - updating `quick-start` terraform and documentation. We now also publish the binary to releases.hashicorp.com, so we can document that as the source for building container images.

### DIFF
--- a/README.md
+++ b/README.md
@@ -97,7 +97,7 @@ curl --silent https://releases.hashicorp.com/vault-lambda-extension/0.4.0/vault-
 unzip vault-lambda-extension.zip
 ```
 
-Optionally, you can verify the integrity of the downloaded zip using the instructions [here](https://www.hashicorp.com/security#:~:text=Release%20Archive%20Checksum%20Verification).
+Optionally, you can verify the integrity of the downloaded zip using the release archive checksum verification instructions [here](https://www.hashicorp.com/security).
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -80,29 +80,28 @@ If you deploy your Lambda function as a zip file, you can add the extension
...
+Fetch the binary from releases.hashicorp.com:
 
 ```bash
-# Requires Golang installed. Run from the root of this repository.
-GOOS=linux GOARCH=amd64 go build -o vault-lambda-extension main.go
+# Requires `curl` and `unzip`
+curl --silent https://releases.hashicorp.com/vault-lambda-extension/0.4.0/vault-lambda-extension_0.4.0_linux_amd64.zip \
+  --output vault-lambda-extension.zip
+unzip vault-lambda-extension.zip
 ```
```
tvoran, line 98: I wonder if it's worth mentioning the [SHA256SUMS file](https://releases.hashicorp.com/vault-lambda-extension/0.4.0/vault-lambda-extension_0.4.0_SHA256SUMS) as an integrity check for the download.

tvoran (approved):

README.md
tomhjp: Good idea, done.

PR #47, opened 2021-08-11
Support performance standbys

backjo:
This PR updates the Vault API to the latest version and adds support for performance standbys. 

Notably, this brings in the `X-Vault-Index` header with client-controlled consistency, which we then leverage to ensure that reads after the initial login write are done consistently.

Today, using performance standbys w/ this extension results in a fairly large number of errors around:
```
error performing token check: failed to persist lease entry: cannot write to readonly storage
```


tomhjp:
Hi @backjo, thanks for raising this! This is a good first step, but the upgrade on its own won't fix any consistency issues because each client needs to handle reading and setting the `X-Vault-Index` header according to its own needs.

For proxy server mode, I believe we shouldn't make any modifications - it's up to the consumer of the proxy to decide what consistency requirements they have.

For writing secrets to disk though, we should definitely support using this feature to ensure our login state has been replicated, in between logging in and reading the pre-configured secrets. Is this where you're running into issues?

We could possibly also wait for index values between each secret read, but I can't think of any actual likely scenarios in which successive secret reads would depend on each other, and in any case we don't make any guarantee about the order in which secrets are read. I would like to avoid adding a dependency between secret reads so that we retain the option to parallelise the secret reads in the case that someone may have a significant number of secrets to read.

So in summary, I think we need to read a `X-Vault-Index` header from the login request response, and if that's present, use it for the following reads that are made for writing to disk, if that makes sense?

backjo:
@tomhjp ah, you're right! When diving into the code, I thought the restore/record state logic was in the core client - but indeed it's in the proxy agent, which makes way more sense!

Yep, writing secrets to disk is where we're running into issues - the read after write (login) is pretty consistently causing the aforementioned error.

main.go
```diff
@@ -107,16 +107,23 @@ func runExtension(ctx context.Context, logger *log.Logger, wg *sync.WaitGroup) (
 		return nil, fmt.Errorf("nil client returned: %w", err)
 	}
 
+	var newState string
+	//Leverage Vault helpers for eventual consistency on login
+	client.VaultClient = client.VaultClient.WithResponseCallbacks(api.RecordState(&newState)).WithRequestCallbacks(api.RequireState(newState))
```
tomhjp, line 112: `newState` is passed to the request callback by value before getting set, so I don't think this will quite work yet. It also asks Vault to do a little more work than it needs to if the extension is configured to e.g. read dynamic secrets, because each dynamic secret will return a new state and the next secret read will wait for that state to be replicated.

Could we only set a response callback on this line, and then clear that response callback and add a request callback on L118 below? Alongside the clearing of all callbacks at L125, I think that will have all the bases covered then.

tomhjp (commented):
Thanks for the update, looking pretty close now. In terms of testing, do you think you might be able to test this out in your environment to confirm? If you have `aws` CLI installed and authenticated, you can run `make publish-layer-version` and it will package and upload the extension and display a new ARN you can set in your Lambda. I'll try to run some of my own testing too, but it would be great to have higher confidence given how complex the setup will be.

tomhjp:
Also just to note, hopefully your next commit should trigger CI, as I've enabled the required setting in CircleCI now.

tomhjp (approved):
LGTM. Thank you for the contribution! I've run it through a bit of my own testing with a modified quick-start guide. We'll run a release that includes this change this week, possibly today.

PR #53, opened 2021-10-25
Update jira action

tomhjp:
* No longer check for Vault team membership
* Tweak jira states and search parameters


benashz (approved):
SGTM!

PR #58, opened 2022-02-09
Add caching support in the local proxy server

tvoran:
(Continuing #52)

Adds caching support to the local proxy server for the Vault lambda extension. To use, set the VAULT_DEFAULT_CACHE_TTL environment variable to a time.Duration value (15m, 1h). Requests with the header "X-Vault-Cache-Control: cache" will be returned from the cache on a hit, and responses will be stored in the cache if not. The cache can be invalidated by setting the header to "recache".

Also if the VAULT_DEFAULT_CACHE_ENABLE environment variable is set to `true`, all requests will behave as though "X-Vault-Cache-Control: cache" was set.

Changes from the original PR:

- [x] Use go-cache for cache management
- [x] Use a hash of the full request as the cache key
- [x] Rename VAULT_CACHE_TTL -> VAULT_DEFAULT_CACHE_TTL
- [x] Use headers instead of query parameters for cache control
- [x] Option to always have caching enabled


internal/proxy/cache.go
```diff
@@ -0,0 +1,205 @@
...
+
+// constructs the CacheKey for this request and token and returns the SHA256
+// hash
+func makeRequestHash(logger *log.Logger, r *http.Request, token string) (string, error) {
+	reqBody, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		return "", fmt.Errorf("failed to read request body: %w", err)
+	}
+	if r.Body != nil {
+		r.Body.Close()
```
tomhjp, line 69: Is it possible that the body also needs closing when there is an error above? And should we also log the returned error from `Close` if it's non-nil?

internal/proxy/proxy.go
```diff
@@ -39,25 +47,65 @@ func proxyHandler(logger *log.Logger, client *vault.Client) func(http.ResponseWr
...
+		if shallFetchCache(fwReq, cache) {
+			// Construct the hash for this request to use as the cache key
+			cacheKeyHash, err = makeRequestHash(logger, r, token)
+			if err != nil {
+				logger.Printf("failed to compute request hash: %s", err)
+				http.Error(w, "failed to read request", http.StatusInternalServerError)
+				return
+			}
+			// Check the cache for this request
+			data, err := cache.Get(cacheKeyHash)
```
tomhjp, line 60: I think we should probably lock an exclusive mutex, and immediately defer the unlock, before we `Get` from the cache. If two requests come in simultaneously, the last one should wait until the first finishes. We could make the mutex specific to the particular key we're retrieving, but that seems likely to be overkill in complexity compared to the typical scale at which a single function will be making Vault requests.

tomhjp (commented):
Looking good! Some questions to discuss around error handling and concurrency, but mostly nits at this stage. I haven't reviewed the tests yet, but will review on the next round.

README.md
```diff
@@ -188,6 +189,14 @@ STS endpoint by setting `AWS_STS_REGIONAL_ENDPOINTS=regional`, because both the
 SDK and Vault IAM auth method default to using the global endpoint in many regions.
 See documentation on [`sts_regional_endpoints`][lambda-sts-regional-endpoints] for more information.
 
+### Caching
+
+Caching can be configured for the extension's local proxy server so that it does not forward every HTTP request to Vault. To turn on caching, set VAULT_DEFAULT_CACHE_TTL to a valid value that is parsable by time.Duration in Go, for example, "15m", "1h", "2m3s" or "1h2m3s", depending on application needs. An invalid or negative value will be treated the same as a missing value, in which case, caching will not be set up and enabled.
```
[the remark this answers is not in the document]
tvoran: Wrapped in b260178c9678c4ab7400662fb053240a055ac2db

internal/proxy/cache.go
```diff
@@ -0,0 +1,205 @@
...
+	c.data.Delete(keyStr)
+}
+
+func copyHeaders(dst, src http.Header) {
+	for k, vs := range src {
+		for _, v := range vs {
+			dst.Add(k, v)
+		}
+	}
+}
```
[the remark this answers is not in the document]
tvoran: Moved in c1dd09aa3922f8560490e0b399478ceb783842a4

internal/proxy/cache.go
```diff
@@ -0,0 +1,205 @@
...
+	options.recache = strutil.StrListContains(values, headerOptionRecache)
+}
+
+func shallFetchCache(r *http.Request, cache *Cache) bool {
+	if cache == nil {
+		return false
+	}
+	cacheControl := r.Header.Get(VaultCacheControlHeaderName)
+	options := &CacheOptions{}
+	setCacheOptions(cacheControl, options)
```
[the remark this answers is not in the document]
tvoran: Yeah, I like that better. Changed in c1dd09aa3922f8560490e0b399478ceb783842a4.

internal/proxy/proxy.go
```diff
@@ -11,10 +11,18 @@ import (
 	"github.com/hashicorp/vault/sdk/helper/consts"
 )
 
+const (
+	VaultIndexHeaderName        = "X-Vault-Index"
```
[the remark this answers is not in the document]
tvoran: I think that makes sense, except for X-Vault-Cache-Control since that's specific to VLE for now.

tvoran:
@JinlianWang Yes that was added in a0256b1.

internal/proxy/cache.go
```diff
@@ -0,0 +1,205 @@
...
+
+func retrieveData(resp *http.Response) *CacheData {
+	data := &CacheData{}
+	data.StatusCode = resp.StatusCode
+	data.Header = resp.Header
+
+	var buf bytes.Buffer
+	_, err := io.Copy(&buf, resp.Body)
+	if err != nil {
+		data.StatusCode = http.StatusInternalServerError // also cache errors
```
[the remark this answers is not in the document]
tvoran: Yeah, I think I agree on this, and good point about the 503s. Will take a look.

internal/proxy/cache.go
```diff
@@ -0,0 +1,205 @@
...
+
+func shallFetchCache(r *http.Request, cache *Cache) bool {
+	if cache == nil {
+		return false
+	}
+	cacheControl := r.Header.Get(VaultCacheControlHeaderName)
+	options := &CacheOptions{}
+	setCacheOptions(cacheControl, options)
+	cacheable := options.cacheable && !options.recache
+	return r.Method == "GET" && cacheable
```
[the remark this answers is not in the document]
tvoran: I have been wondering this too, and I suspect you're right that other methods should be ok to be cached. It'll just need a little more testing and investigation.

internal/proxy/proxy.go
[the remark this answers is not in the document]
tomhjp: Yeah agreed 👍

internal/proxy/proxy.go
tomhjp: Yeah the underlying cache access looks safe. My thinking was more at the application behaviour level. If we don't lock between a cache miss and the underlying request being made, then we could have multiple in-flight Vault requests for the same secret simultaneously. And that could also result in inconsistent results from the cache. e.g. I think currently [this](https://mermaid-js.github.io/mermaid-live-editor/edit#pako:eNp9kU-LwjAQxb9KmLNFdNVDDkLriteFXfaUy5hMtWASTROwFL-7o1S07tLb8H7vvfyZFrQ3BBJqOiVymj4r3AW0yqGOPoj8MRTK5dly-RX8uZFis_4RZjsuvVeu-F--S4x-MR2iFPkgLfr0rmbPXsapppBNph_PaMdWqPf0BzPvhd4b8w7O5ouhxhfMvBcafPbL8WwWZfBW6FstjMBSsFgZ_vNWOSEUxD1ZUiB5NFTeLqpAuQtb09FgpLWpeAMgSzzUNAJM0X83ToOMIdHD1O2tc12uGiidPA) could happen, but we want [this](https://mermaid-js.github.io/mermaid-live-editor/edit/#pako:eNqFkMEKwjAQRH8l7LlF1FsOhVbFq6B4ymVNtrbQJJomoEj_3VQqRax4W-bNg2QeIK0i4NDSNZCRtK7x7FALg9Jbx_L3UQiTp1m2c_Z252y7OTB1mpXWClNMx68ooiOGxnOWf9JXmo5mxKEll84Xy1Ed2AplRV848gkppsWYstJZzWTv_3nnTw8S0OQ01iqO9BCGMQG-Ik0CeDwVlf1PBAjTxWq4KPS0UXWcDHiJTUsJYPB2fzcSuHeB3qVh6KHVPQHhd4Nj) to happen instead. The diagram has two actors, but they could just represent different threads within the same process.

internal/proxy/cache.go
```diff
@@ -0,0 +1,199 @@
...
+const (
+	// Return a cached response if it exists, otherwise fall back to Vault and
+	// cache the response
+	headerOptionCacheable = "cache"
+
+	// Send the request to Vault and cache the response
+	headerOptionRecache = "recache"
+
+	// Ignore the cache and send the request to Vault, do not cache the response
+	headerOptionNocache = "nocache"
```
calvn, line 31: The [standard cache-control headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control) use a slightly different directive for these two behavior. `no-cache` is equivalent to `recache` and `no-store` is equivalent to no cache. These are vault-specific cache headers though, so I'll leave it up to you on whether we want to use the same naming convention.

internal/proxy/cache.go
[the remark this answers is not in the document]
calvn: I agree as well. On Agent we [only cache 2xx (successful) responses](https://github.com/hashicorp/vault/blob/eefc9817a1147a877d77dc1700a29cab830b83a6/command/agent/cache/lease_cache.go#L283-L287).

internal/proxy/proxy.go
```diff
@@ -7,14 +7,24 @@ import (
...
 
+	"github.com/hashicorp/vault-lambda-extension/internal/config"
 	"github.com/hashicorp/vault-lambda-extension/internal/vault"
 	"github.com/hashicorp/vault/sdk/helper/consts"
 )
 
+const (
+	VaultCacheControlHeaderName = "X-Vault-Cache-Control"
+
+	// TODO: remove these definitions once they're available in Vault's api package
```
calvn, line 18: Does this need to be addressed before this PR is merged?

internal/proxy/proxy.go
calvn: We need a way to track inflight requests in order to avoid the situation that Tom described above, where there might be two concurrent inflight requests leading to two of the same requests (e.g. a creds generation request) to be made against the Vault server. 

A global exclusive mutex lock would make the entire caching mechanism synchronous and it'd only be able to process one request at a time. The[ approach that I took for agent](https://github.com/hashicorp/vault/pull/10705/files) to address this issue is to make a second and separate inflight cache that is clears once the first request "wins". The other requests will block (using a channel) until then. This is a way to coordinate concurrent inflight requests of the same type without having to manage locks per request or making the entire cache operate on a global lock.

internal/proxy/proxy.go
tvoran: I don't believe so.

internal/proxy/cache.go
tvoran: I'm not really sure the body needs closing when there is an error above, since an error here will stop processing the request in the calling handler. That being said I added more logging and another call to r.Body.Close() out of an abundance of caution in 73bfb8c815700f7a4bbe6951eae49d7c95579244.

internal/proxy/cache.go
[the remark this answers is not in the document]
tvoran: I refactored the cache refresh portion in 6f639320ba45d499b1b491be433a4a11ff3b8d06 to only cache 2xx responses.

internal/proxy/proxy.go
tomhjp: I think allowing multiple inflight requests is more complexity than we need. The scale demands on this proxy are much smaller than for Agent, because each instance of the proxy is dedicated to only one instance of the Lambda function at a time. Given that, I actually don't think a global mutex would be disastrous, but I'm happy to start with a mutex per cache key instead.

internal/proxy/cache.go
tomhjp: Yeah sorry, to be clear, I actually wasn't sure of the answers to my questions. As you say, handling those cases is showing an abundance of caution, but LGTM 👍

internal/proxy/proxy.go
tvoran: Appreciate the discussion and the diagrams especially here. I added a mutex per cache key in 6e523bf, and then moved the lock acquisition to the top of the proxyHandler in 8c49e18.

I was thinking about a similar scenario with two requests, but one of them has "recache" set. The "recache" request would skip the cache.Get() (and the locking), but still get the cred from Vault and store it in the cache.

internal/proxy/proxy.go
tvoran: Added them to Vault's api package in https://github.com/hashicorp/vault/pull/14067

internal/proxy/cache_test.go
```diff
@@ -0,0 +1,498 @@
...
+	r.Header.Set(VaultCacheControlHeaderName, headerOptionCacheable)
+	statusCode := 200
+	body := "Hello World"
+	resp := &http.Response{
+		Status:        "200 OK",
+		StatusCode:    statusCode,
+		Proto:         "HTTP/1.1",
+		ProtoMajor:    1,
+		ProtoMinor:    1,
+		Body:          ioutil.NopCloser(bytes.NewBufferString(body)),
```
tomhjp, line 491: nit: `retrieveData` shouldn't touch the request body, so it would be a slightly tighter test if this is nil or a different body.

internal/proxy/cache_test.go
```diff
@@ -0,0 +1,498 @@
...
+				},
+			},
+			RequestBody: nil,
+		}
+		cacheKeyHash, err := computeRequestID(&cacheKey)
+		require.NoError(t, err)
+		require.NotEmpty(t, cacheKeyHash)
+		cache.Set(cacheKeyHash, cacheData)
+
+		time.Sleep(5 * time.Second)
```
tomhjp, line 306: Can we make the expiry and sleep time something like 10ms and 20ms respectively? The cache package's own tests use times of that order of magnitude, and looking at the implementation, I don't think it will be flaky.

tomhjp (approved):
Only minor comments - I think this looks pretty much ready to go 👍

README.md
```diff
@@ -170,6 +170,8 @@ Environment variable    | Description | Required | Example value
 `VAULT_TLS_SERVER_NAME` | Name to use as the SNI host when connecting via TLS | No | `vault.example.com`
 `VAULT_RATE_LIMIT`      | Only applies to a single invocation of the extension. See [Vault Commands (CLI)][vault-env-vars] documentation for details. Ignored by proxy server | No | `10`
 `VAULT_NAMESPACE`       | The namespace to use for pre-configured secrets. Ignored by proxy server | No | `education`
+`VAULT_DEFAULT_CACHE_TTL` | The time to live configuration (aka, TTL) of the cache used by proxy server. Must have a unit and be parsable by time.Duration. | No | `15m`
+`VAULT_DEFAULT_CACHE_ENABLED` | Enable caching for all requests, without needing to set the X-Vault-Cache-Control header for each request. Must be set to a boolean value. | No | `true`
```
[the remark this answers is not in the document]
tvoran: Agreed; I added mention of it being required in 083f08b73f2e1b1638521519fe1d3aac0d08969a.

README.md
```diff
@@ -188,6 +190,36 @@ STS endpoint by setting `AWS_STS_REGIONAL_ENDPOINTS=regional`, because both the
...
+You can always override caching and forward a request to Vault, for example in
+the case of expired secrets, set "X-Vault-Cache-Control: recache". The proxy
+server will forward the request to Vault and refresh its cache with the response
+so that a later cache hit can retrieve the newer secrets.
+
+Caching may also be enabled by default by setting the environment variable
+`VAULT_DEFAULT_CACHE_ENABLE` to `true`. Then all requests will be cached/fetched
+as though the header 'X-Vault-Cache-Control: cache' was set. Setting the
+'X-Vault-Cache-Control' header to 'nocache' or 'recache' will modify caching
+behavior as described above.
```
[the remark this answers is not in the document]
tvoran: Good point! I reworked this section a bit in 083f08b73f2e1b1638521519fe1d3aac0d08969a to make it flow a little better with fewer surprises.

internal/proxy/cache.go
```diff
@@ -0,0 +1,203 @@
...
+)
+
+type Cache struct {
+	data    *gocache.Cache
+	timeout time.Duration
+	enabled bool
```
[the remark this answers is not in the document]
tvoran: Yep, `defaultOn` is definitely clearer. Changed in bd90b345c483bfb20438482fa489a3ef54380bfc.

internal/proxy/cache_test.go
tvoran: Done in bd90b345c483bfb20438482fa489a3ef54380bfc.

internal/proxy/cache.go
```diff
@@ -0,0 +1,203 @@
...
+	// Send the request to Vault and cache the response
+	headerOptionRecache = "recache"
+
+	// Ignore the cache and send the request to Vault, do not cache the response
+	headerOptionNocache = "nocache"
+)
+
+type Cache struct {
+	data    *gocache.Cache
+	timeout time.Duration
```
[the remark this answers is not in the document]
tvoran: Yep! Gone in bd90b345c483bfb20438482fa489a3ef54380bfc.

internal/proxy/cache_test.go
tvoran: Cool, changed it to 10ms and 20ms in bd90b345c483bfb20438482fa489a3ef54380bfc and it seems to be working fine.

internal/proxy/cache.go
tvoran: That's a good point, though I do think the headers as-is are clearer as to their meaning than those of the standard. However I don't like that there is overlap with the standard headers, since that could lead to confusion, so I'm happy to change them if we just want to stick with the standard.

calvn: That's fair, these are not strictly the same standard cache-control headers.

internal/proxy/cache.go
```diff
@@ -0,0 +1,203 @@
...
+
+func setupCache(cacheConfig config.CacheConfig) *Cache {
+	if cacheConfig.TTL <= 0 {
+		return nil
+	}
+	return NewCache(cacheConfig)
+}
+
+func parseCacheOptions(cacheControlHeader string) *CacheOptions {
+	values := strings.Split(cacheControlHeader, ",")
```
calvn, line 163: If we support passing in more that one caching option (with the most restrictive one winning), then we should also parse for the case we're `X-Vault-Cache-Control` is passed more than once since that's [equivalent](https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2) to passing a comma-separated list

calvn (approved):
One small comment around header parsing, otherwise this looks great!

internal/proxy/cache.go
tvoran: Yep, that's a good point, added the extra parsing in 675aa462eade8a80299d379d8b89a097d19de39f.

Issue #31, opened 2021-04-19
Lambda function invoked before credentials retrieved/written to disk


armandAkop:
Lambda Runtime: Node 14x

Hi, I've noticed some errors when trying to read my vault secret with path `/tmp/vault/secret.json`. This seems to only happen when the lambda function hasn't been invoked in a while (30+ minutes). Once the lambda fails, the coordinating step function retries and it is successful the second time around. Here is the initial error from the cloudwatch logs. 

Notice: "authenticating to Vault", then it seems to jump straight to the code which tries to read the secret.

```

  | 2021-04-19T10:20:19.365-07:00 | [vault-lambda-extension] 2021/04/19 17:20:19 Initialising
-- | -- | --
  | 2021-04-19T10:20:19.365-07:00 | [vault-lambda-extension] 2021/04/19 17:20:19 authenticating to Vault
  | 2021-04-19T10:20:20.134-07:00 | 2021-04-19T17:20:20.133Z undefined ERROR Uncaught Exception {"errorType":"Runtime.ImportModuleError","errorMessage":"Error: Cannot find module '/tmp/vault/secret.json'\nRequire stack:\n- /var/task/services/secretReader.js\n
  | 2021-04-19T10:20:20.147-07:00 | EXTENSION Name: vault-lambda-extension State: Registered Events: [INVOKE,SHUTDOWN]
```

Then on the following **successful** invocation...

```
  | 2021-04-19T10:20:20.462-07:00 | [vault-lambda-extension] 2021/04/19 17:20:20 Initialising
-- | -- | --
  | 2021-04-19T10:20:20.462-07:00 | [vault-lambda-extension] 2021/04/19 17:20:20 authenticating to Vault
  | 2021-04-19T10:20:21.198-07:00 | [vault-lambda-extension] 2021/04/19 17:20:21 Initialised
  | 2021-04-19T10:20:21.198-07:00 | [vault-lambda-extension] 2021/04/19 17:20:21 Waiting for event...
  | 2021-04-19T10:20:21.198-07:00 | [vault-lambda-extension] 2021/04/19 17:20:21 Starting HTTP server
  | 2021-04-19T10:20:23.056-07:00 | EXTENSION Name: vault-lambda-extension State: Ready Events: [INVOKE,SHUTDOWN]
  | 2021-04-19T10:20:23.057-07:00 | [vault-lambda-extension] 2021/04/19 17:20:23 Received event
[...]

PR #61, opened 2022-02-28
Delay extension registration (hence Lambda runtime initialization) un…

JinlianWang:
…til extension is fully initialized. Effort to fix issue: https://github.com/hashicorp/vault-lambda-extension/issues/31


JinlianWang:
Close it until we test it more internally. Will re-open afterwards.

JinlianWang:
Reopen this pull request after a bit more testing. More testing would be great and will be done, though we do want to move this fix along promptly in order to make it into next release.

tomhjp (approved):
Thanks for digging into this and raising it. I had a test by adding some big `time.Sleep`s and verified that it does indeed reliably work! 🎉 

This is a great one to clear up. Hope you don't mind I just merged in main and added a changelog to save a review cycle. Thanks for the contribution!

PR #62, opened 2022-03-02
Update go to 1.17.7

tvoran:
Updates the version of golang used in circleci jobs and Makefile to 1.17.7. Also updates the go.mod go version to 1.17. Removes the working_directory directive from the circleci config since the new `cimg/go` image seems to have a little different layout than the old `circleci/golang` image.


go.mod
```diff
@@ -1,14 +1,61 @@
 module github.com/hashicorp/vault-lambda-extension
 
-go 1.15
+go 1.17
```
tomhjp, line 3: Does this gate the format update below where direct and indirect dependencies are split up? This setting only specifies the minimum version of Go required to build the module, so I'm surprised to see it updated, but I don't have any problem with enforcing recent versions, especially as this is an application and not a library.

tomhjp (approved):

go.mod
tvoran: Yeah exactly, running `go mod tidy` after updating this line to 1.17 produces this new format for go.mod, I think to support [module graph pruning](https://go.dev/doc/go1.17#go-command).

I went ahead and bumped it since as you mentioned this is an application and not a library that other applications would be expected to import.

Issue #59, opened 2022-02-15
Usability: Can we trim environment variables?


AdrienneCohea:
We had a very interesting thing happen today. We're using `VAULT_AUTH_ROLE` but missed a trailing space, and as a result got a very vexacious

`Error: entry for role rolename  not found`

We checked and rechecked our Vault AWS roles in the auth method and our IAM role multiple times, but missed the extra space in the error format string. And yeah, AWS could have helped us out here by trimming the spaces in the text box for us.

But I wonder, could we trim the environment variables here? https://github.com/hashicorp/vault-lambda-extension/blob/8071362e261d315bd6a63b183494ecfdef336cc6/internal/config/auth.go#L27

AdrienneCohea:
Curiously Github took out the extra space in my example of the error message. But you get the idea: VAULT_AUTH_ROLE was "rolename " rather than "rolename"

tomhjp:
Hi @AdrienneCohea, thanks for raising this! Seems like a completely reasonable improvement, especially if the console UI is making it easy to misconfigure. Happy to take PRs, or otherwise I'll try to get it in for the next release, which should be coming up relatively soon.

PR #63, opened 2022-03-03
Use levelled logger

### LINKED ISSUE #57: Possible to have a setting to control logging level?
Logging may be helpful during debugging/development, though would be nice to be able to turn off some logging if end users choose to.

tomhjp:
Closes #57, #59

Switch to using `hclog` so that we can have levels, and use the same environment variable that Vault uses to control it. Comments welcome on the levels I've chosen for each log line.

### DIFF
--- a/internal/proxy/proxy.go
+++ b/internal/proxy/proxy.go
@@ -4,10 +4,10 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"log"
 	"net/http"
 	"net/url"
 
+	"github.com/hashicorp/go-hclog"
 	"github.com/hashicorp/vault-lambda-extension/internal/config"
 	"github.com/hashicorp/vault-lambda-extension/internal/vault"
 	"github.com/hashicorp/vault/sdk/helper/consts"
@@ -19,7 +19,7 @@ const (
 )
 
 // New returns an unstarted HTTP server with health and proxy handlers.
-func New(logger *log.Logger, client *vault.Client, cacheConfig config.CacheConfig) *http.Server {
+func New(logger hclog.Logger, client *vault.Client, cacheConfig config.CacheConfig) *http.Server {
 	cache := setupCache(cacheConfig)
 	mux := http.ServeMux{}
 	mux.HandleFunc("/", proxyHandler(logger, client, cache))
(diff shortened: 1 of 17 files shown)


main.go
```diff
@@ -27,21 +27,36 @@ import (
...
 	extensionName = "vault-lambda-extension"
+	vaultLogLevel = "VAULT_LOG_LEVEL" // Optional, one of TRACE, DEBUG, INFO, WARN, ERROR, OFF
 )
 
 func main() {
-	logger := log.New(os.Stderr, fmt.Sprintf("[%s] ", extensionName), log.Ldate|log.Ltime|log.LUTC)
+	hclog.LevelFromString(os.Getenv(vaultLogLevel))
+	logger := hclog.New(&hclog.LoggerOptions{
+		Level: hclog.Debug,
+	})
```
tvoran, line 37: Looks like we need to use the result from `hclog.LevelFromString()` to set the level,  and default to info if its not set? Something like:

```suggestion
	level := hclog.LevelFromString(os.Getenv(vaultLogLevel))
	logger := hclog.New(&hclog.LoggerOptions{
		Level: hclog.Info,
	})
	if level != hclog.NoLevel {
		logger.SetLevel(level)
	}
```

tomhjp: Doh, sorry that was very sloppy.

tomhjp: Fixed - `hclog` handles switching to the default level if `NoLevel` is set internally.

internal/proxy/proxy.go
```diff
@@ -99,7 +99,7 @@ func proxyHandler(logger *log.Logger, client *vault.Client, cache *Cache) func(h
 
 		if doCacheSet && resp.StatusCode < 300 {
 			cache.Set(cacheKeyHash, retrieveData(resp, respBody))
-			logger.Printf("Refreshed cache for: %s %s", r.Method, r.URL.Path)
+			logger.Debug("Refreshed cache for: %s %s", r.Method, r.URL.Path)
```
tvoran, line 102:
```suggestion
			logger.Debug(fmt.Sprintf("Refreshed cache for: %s %s", r.Method, r.URL.Path))
```

tvoran (approved):

PR #67, opened 2022-03-22
Use CRT for releases and CI

tvoran:
Release version is specified when manually triggering the build workflow. Go version is determined using the .go-version file. Will build for arm64 as well as amd64.

TODO:

- [X] Run CI tests in a GitHub workflow
- [X] Setup CRT for binary releases
- ~[ ] Setup AWS layer releases in a GitHub workflow~ (this will be done in another repo)


.github/workflows/test.yml
```diff
@@ -0,0 +1,87 @@
...
+            echo "Git directory has go.mod or go.sum changes"
+            exit 1
+          fi
+
+  test:
+    needs:
+      - get-go-version
+    runs-on: ubuntu-latest
+    env:
+      TEST_RESULTS_DIR: /tmp/test-results
```
tomhjp, line 63: I think we can probably remove this part - I'm not sure GHA has an equivalent feature for storing test artifacts like we do in CircleCI.

.release/ci.hcl
```diff
@@ -0,0 +1,120 @@
...
+
+project "vault-lambda-extension" {
+  team = "vault"
+  slack {
+    notification_channel = "CRF6FFKEW"
+  }
+  github {
+    organization = "hashicorp"
+    repository = "vault-lambda-extension"
+    release_branches = ["main", "VAULT-5260/use-crt"]
```
tomhjp, line 11: Leaving a comment as a reminder we should remove the test branch before merging.

.github/workflows/build.yml
```diff
@@ -0,0 +1,96 @@
...
+
+  build:
+    needs:
+      - get-go-version
+      - get-product-version
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        goos: [linux]
+        goarch: ["arm", "arm64", "386", "amd64"]
```
tomhjp, line 70: I think we should limit this to `arm64` and `amd64` as I don't think AWS supports the others and we won't be testing using them in any way

.github/workflows/build.yml
```diff
@@ -0,0 +1,96 @@
...
+          go-version: ${{ needs.get-go-version.outputs.go-version }}
+
+      - name: Build
+        env:
+          GOOS: ${{ matrix.goos }}
+          GOARCH: ${{ matrix.goarch }}
+        run: |
+          make build
+          mkdir out
+          zip -r -j out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip pkg/extensions
```
tomhjp, line 91:
```suggestion
          cd pkg/ && zip -r out/${{ env.PKG_NAME }}_${{ needs.get-product-version.outputs.product-version }}_${{ matrix.goos }}_${{ matrix.goarch }}.zip extensions && cd ..
```

I haven't tested it out recently, but I think AWS Lambda requires the binary to be in the extensions folder.

tomhjp (commented):
Looking good. FWIW I would support leaving the final layer publish step in CircleCI for now. Rel-eng have a ticket to track adding support for publishing AWS Lambda layers, so switching directly to that in the future might be an easier option for us.

.github/workflows/build.yml
tvoran: The security scan fails when the zip contains directories: https://github.com/hashicorp/crt-workflows-common/runs/5652327667?check_suite_focus=true

.github/workflows/build.yml
tvoran: Changed in a54b140

tomhjp (approved):
Looks like this is ready to go if testing goes smoothly 👍

tvoran:
Tested the arm64 binary, and it seems to work, so I think we're good here 👍

PR #79, opened 2022-08-04
quick-start: Use templatefile instead of template module

swenson:
The `template` module is deprecated and is not supported for darwin/amd64. Luckily, our use of it is easily replaced by the `templatefile` function.


tomhjp (approved):

benashz (approved):
LGTM!

swenson:
Thanks!

PR #96, opened 2023-02-21
update x/net package to resolve security vulnerability

raymonstah:


tvoran (approved):

tvoran:
Looks like this sort of update gets mentioned in the changelog too, so might want to add some entries there.

PR #145, opened 2024-08-18
Exclude distributed tracing headers from cache key

### LINKED ISSUE #146: Distributed Tracing causing Unique Cache Keys
I noticed that if I run this extension alongside a layer that provides distributed tracing instrumentation of my lambda function, then my requests always end up missing the cache and going to network. Distributed tracing relies on adding headers like traceparent which contain unique trace identifiers, so it makes sense that the cache key would be different on each request. Unfortunately that means lots of longer network calls and also a needlessly bloated cache in memory if you didn't realize this was happening and made all requests cache able.

aztalbot:
- When using distributed tracing layers for New Relic or Open Telemetry, all outbound requests get decorated with standard trace context headers
- If using the vault extension alongside one of those distributed tracing layers, caching within the proxy won’t work as expected because, traceparent header, for example contains randomly generated IDs, rendering each request unique based on the existing cache key logic
- This PR changes the cache to ignore those standard tracing headers

fixes #146

### DIFF
--- a/internal/proxy/cache.go
+++ b/internal/proxy/cache.go
@@ -120,8 +120,8 @@ func computeRequestID(key *CacheKey) (string, error) {
 	cloned.Header.Del(VaultCacheControlHeaderName)
 	// remove standard distributed tracing headers (https://www.w3.org/TR/trace-context/),
 	// otherwise instrumented requests will always be unique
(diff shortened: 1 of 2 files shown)


fairclothjm (approved):
Thanks!

PR #150, opened 2024-08-29
Strip Monotonic Clock Fixes #113

### LINKED ISSUE #113: Token refresh issue with Snapstart and vault-lambda-extension layer
Hi,
We are using the arn:aws:lambda:${aws:region}:634166935893:layer:vault-lambda-extension:13 layer for connecting to our Enterprise vault.
For improving the cold start we enabled snapstart on the published versions.
Seems the extension layer is not refreshing the token after its expiry or after few hours and we are getting Forbidden errors.
**ERROR RouteExecutor:444 - Unexpected error occurred: Client 'http://127.0.0.1:8200/v1/<Path>data': Forbidden**
The initial ttl value was set to 1d as well tried with 1h. Still the issue persists.

The snapstart takes the snapshot of the Init phase environment and caches it and is reused every time when there is a new instance getting spinned up.
We are not sure if the vault layer is initialising in the init phase and is not able to refresh the connections post sometime.
Could you help fixing the issue.

Thanks,
Nithya

wettowelreactor:
When the extension is suspended by Lambda, such as when using SnapStart, the monotonic clock is also suspended. This leads to errors when calculating Token TTL. Striping the monotonic clock from the timestamps by using `Round(0)` forces the use of wall clock mitigating this issue.

### DIFF
--- a/internal/vault/client.go
+++ b/internal/vault/client.go
@@ -83,7 +83,7 @@ func NewClient(name, version string, logger hclog.Logger, vaultConfig *api.Confi
 
 // Token synchronously renews/re-auths as required and returns a Vault token.
 func (c *Client) Token(ctx context.Context) (string, error) {
-	start := time.Now()
+	start := time.Now().Round(0)
 	c.logger.Debug("fetching token")
 	c.mtx.Lock()
 	defer c.mtx.Unlock()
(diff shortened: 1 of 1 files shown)


wettowelreactor:
#113

fairclothjm (approved):
Thanks!

PR #165, opened 2025-02-20
fix security scanner configuration

hashiblaum:
fixes security scanner configuration location


jaireddjawed (approved):

.release/security-scan.hcl
```diff
@@ -2,9 +2,20 @@
 # SPDX-License-Identifier: MPL-2.0
 
 binary {
-	secrets      = true
-	go_modules   = true
-	osv          = true
-	oss_index    = false
-	nvd          = false
+  secrets      = true
+  go_modules   = true
```
tvoran, line 6: We might want to also add `go_stdlib = true` to catch issues like [GO-2025-3420](https://pkg.go.dev/vuln/GO-2025-3420).

PR #167, opened 2025-03-04
Update repo with new layer versions

hashiblaum:
Update README and quick-start terraform with new layer versions from 0.12.0 release

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,13 +18,13 @@ Lambda function, depending on your desired architecture.
 amd64 (x86_64):
 
 ```text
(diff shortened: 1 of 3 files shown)


jaireddjawed (dismissed):

tvoran (dismissed):
Looks like we often bump the quick-start vault version around release time too: https://github.com/hashicorp/vault-lambda-extension/blob/6d6dd8a78d134c5b1143a4863f7fcbf097cc2944/quick-start/terraform/variables.tf#L16

tvoran (approved):

PR #191, opened 2026-05-05
[VAULT-22277] AWS SDK Migration to Version 2

bcsamrudh:
**Description**
This pull request migrates the project from the legacy AWS SDK for Go (aws-sdk-go) to the modern AWS SDK for Go v2 (aws-sdk-go-v2). This update affects the main application code, tests, and dependencies, and brings the codebase up to date with current AWS SDK.

**Key Changes:**

1.  Replaced all usage of `github.com/aws/aws-sdk-go` with `github.com/aws/aws-sdk-go-v2` and its submodules throughout the codebase, including main logic, tests, and supporting modules. Refactored how AWS sessions, credentials, and service clients are initialized and used.

2.  Vault AWS authentication flows were updated to use SDK v2 configuration and credential provider patterns, including STS client usage and request-signing behavior changes required by v2 APIs.

3. Updated the Dockerfile to use the latest available base image and dependencies to ensure compatibility with the AWS SDK v2 migration.Additionally, this PR changes the default EC2 instance type in the quick-start Terraform configuration.

4. Client now uses constructor-provided authConfig (c.authConfig) instead of Config from the environment (config.AuthConfigFromEnv())

5. Previously, the configuration would always pick the latest Ubuntu AMI, including ARM-based variants. Since the instance type uses the x86_64 architecture, this could result in errors such as:
" Error: creating EC2 Instance: InvalidParameterValue: The architecture 'x86_64' of the specified instance type does not match the architecture 'arm64' of the specified AMI. "
This change ensures that only x86 Ubuntu AMIs are selected, keeping the AMI architecture aligned with the instance architecture.

6. Testing: Verified stability through unit tests, build checks, and AWS acceptance tests for all updated functionality.

7.  Code Cleanup: Removed deprecated utilities and updated helper functions to use supported Go and AWS SDK v2 methods.

Ticket : [VAULT-22277](https://hashicorp.atlassian.net/browse/VAULT-22277)

**Testing Evidence:**

1. End to End Testing
tempImageqxFswR

2.  Integration Tests
tempImagebvEoUT
tempImageUS4JH2

3. Unit Tests 
tempImageammgsF

[VAULT-22277]: https://hashicorp.atlassian.net/browse/VAULT-22277?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/proxy/proxy_test.go
```diff
@@ -63,10 +65,11 @@ var (
 func TestProxy(t *testing.T) {
 	fakeVault := fakeVault()
 	defer fakeVault.Close()
-	ses := session.Must(session.NewSession())
-	sts := ststest.FakeSTS(ses)
+	awsCfg, err := awsconfig.LoadDefaultConfig(context.TODO())
+	require.NoError(t, err, "failed to load AWS config ")
```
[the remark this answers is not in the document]
stuti-sr: ignore this one

quick-start/terraform/variables.tf
```diff
@@ -18,7 +18,7 @@ variable "vault_zip_file" {
 
 # Instance size
 variable "instance_type" {
-  default = "t2.micro"
+  default = "t3.micro"
 }
```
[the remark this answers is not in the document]
bcsamrudh: Have updated the PR Description

test/docker-compose.yaml
```diff
@@ -29,9 +29,12 @@ services:
...
-    image: docker.mirror.hashicorp.services/vault:1.6.2
+    image: hashicorp/vault:1.21.1
     command: vault server -dev -log-level=err
     environment:
       - VAULT_DEV_ROOT_TOKEN_ID=root
-      - VAULT_DEV_LISTEN_ADDRESS=0.0.0.0:8200
\ No newline at end of file
+      - VAULT_DEV_LISTEN_ADDRESS=0.0.0.0:8200
+    cap_add:
+      - IPC_LOCK
```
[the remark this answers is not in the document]
bcsamrudh: Have updated the PR Description to include this change.

test/lambda/Dockerfile
```diff
@@ -1,9 +1,9 @@
 # Copyright IBM Corp. 2020, 2025
 # SPDX-License-Identifier: MPL-2.0
 
-FROM docker.mirror.hashicorp.services/vault:1.6.2 as vault
+FROM docker.io/hashicorp/vault:1.21.1 AS vault
 
-FROM docker.mirror.hashicorp.services/golang:1.15-alpine3.13 as builder
+FROM docker.mirror.hashicorp.services/golang:1.25.7-alpine AS builder
```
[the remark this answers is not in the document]
bcsamrudh: Have updated the PR Description to include this change.

test/api/Dockerfile
```diff
@@ -1,7 +1,7 @@
 # Copyright IBM Corp. 2020, 2025
 # SPDX-License-Identifier: MPL-2.0
 
-FROM docker.mirror.hashicorp.services/golang:1.15-alpine3.13 as builder
+FROM docker.mirror.hashicorp.services/golang:1.25.7-alpine AS builder
```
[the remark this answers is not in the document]
bcsamrudh: Have updated the PR Description to include this change.

internal/vault/client.go
```diff
@@ -198,6 +213,41 @@ func (c *Client) login(ctx context.Context) error {
...
+	if region == "" {
+		region = defaultSTSRegion
+	}
+
+	endpoint, err := resolver.ResolveEndpoint(ctx, sts.EndpointParameters{
+		Region:            aws.String(region),
+		UseDualStack:      aws.Bool(options.EndpointOptions.UseDualStackEndpoint == aws.DualStackEndpointStateEnabled),
+		UseFIPS:           aws.Bool(options.EndpointOptions.UseFIPSEndpoint == aws.FIPSEndpointStateEnabled),
+		Endpoint:          options.BaseEndpoint,
+		UseGlobalEndpoint: aws.Bool(false),
```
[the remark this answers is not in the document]
bcsamrudh: UseGlobalEndpoint is set to false to force it to use RegionalEndpoint

internal/vault/client.go
```diff
@@ -126,59 +128,72 @@ func (c *Client) login(ctx context.Context) error {
...
 	}
 
-	headers, err := json.Marshal(req.HTTPRequest.Header)
+	creds, err := stsOptions.Credentials.Retrieve(ctx)
 	if err != nil {
 		return err
 	}
+	if err := v4.NewSigner().SignHTTP(ctx, creds, httpReq, bodyHash, "sts", stsRegion, time.Now()); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
bcsamrudh: Fixed it

internal/vault/client.go
```diff
@@ -42,7 +42,8 @@ type Client struct {
 	VaultConfig *api.Config
 
 	logger     hclog.Logger
-	stsSvc     *sts.STS
+	awsCfg     aws.Config
+	stsSvc     *sts.Client
```
arjunkshashicorp: `stsSvc` is redundant as a struct field — it's just a pre-computed `sts.NewFromConfig(awsCfg)`, and `awsCfg` is already stored on the struct. Since login() is only called on token expiry/revocation (not in the hot path), we can drop the field and create the STS client on-demand inside login() instead. This simplifies the struct and removes a piece of state that doesn't need to live there.

You will need to update the tests after this change

internal/vault/client.go
```diff
@@ -126,64 +128,77 @@ func (c *Client) login(ctx context.Context) error {
 		c.logger.Debug(fmt.Sprintf("Trying to assume role with arn of %s to authenticate with Vault", roleToAssumeArn))
 		sessionName := "vault_auth"
 
-		result, err := c.stsSvc.AssumeRole(&sts.AssumeRoleInput{
-			RoleArn:         &roleToAssumeArn,
-			RoleSessionName: &sessionName,
+		result, err := c.stsSvc.AssumeRole(ctx, &sts.AssumeRoleInput{
```
arjunkshashicorp: `result` is too generic for what it holds. Since only `result.Credentials` is ever accessed, a name like assumeRoleOutput (matching the AWS SDK v2 type `*sts.AssumeRoleOutput` or assumedRole would make the intent immediately clear without needing to trace back to the AssumeRole call.

internal/vault/client.go
```diff
@@ -198,6 +213,41 @@ func (c *Client) login(ctx context.Context) error {
 	return c.updateTokenMetadata(secret)
 }
 
+// resolvedSTSEndpointURL resolves the concrete STS endpoint using the SDK's
+// endpoint resolver, then normalizes it to a URL safe for signing.
+func resolvedSTSEndpointURL(ctx context.Context, stsSvc *sts.Client) (string, error) {
```
arjunkshashicorp: Nit: the function name reads like a past-participle adjective rather than a verb. Go convention prefers imperative names — `resolveSTSEndpointURL` would be more idiomatic.

internal/vault/client.go
```diff
@@ -126,64 +128,77 @@ func (c *Client) login(ctx context.Context) error {
...
+		stsSvc = sts.NewFromConfig(assumedRoleCfg)
 	}
 
-	// ignore out
-	req, _ := stsSvc.GetCallerIdentityRequest(&sts.GetCallerIdentityInput{})
-	req.SetContext(ctx)
-
-	if c.authConfig.IAMServerID != "" {
-		req.HTTPRequest.Header.Add("X-Vault-AWS-IAM-Server-ID", c.authConfig.IAMServerID)
+	stsOptions := stsSvc.Options()
```
arjunkshashicorp, line 185: ### Extract into a `buildIAMAuthPayload` helper

This block does two distinct things — builds and signs a `GetCallerIdentity` request, then packages it into the Vault IAM auth payload. Extracting it makes `login()` easier to follow and the signing logic independently testable.

```go
d, err := buildIAMAuthPayload(ctx, stsSvc, authConfig)
if err != nil {
    return err
}
```

With the helper:

```go
// buildIAMAuthPayload builds and signs a GetCallerIdentity request, then packages
// it into the payload expected by Vault's AWS IAM auth login endpoint.
func buildIAMAuthPayload(ctx context.Context, stsSvc *sts.Client, authConfig config.AuthConfig) (map[string]interface{}, error) {
    stsOptions := stsSvc.Options()
    stsRegion := stsOptions.Region
    if stsRegion == "" {
        stsRegion = defaultSTSRegion
    }
    stsEndpoint, err := resolvedSTSEndpointURL(ctx, stsSvc)
    if err != nil {
        return nil, fmt.Errorf("failed to resolve STS endpoint URL: %w", err)
    }

    body := "Action=GetCallerIdentity&Version=2011-06-15"
    bodyHash := fmt.Sprintf("%x", sha256.Sum256([]byte(body)))

    httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, stsEndpoint, strings.NewReader(body))
    if err != nil {
        return nil, fmt.Errorf("failed to build STS GetCallerIdentity request: %w", err)
    }
    httpReq.Header.Set("Content-Type", "application/x-www-form-urlencoded")
    if authConfig.IAMServerID != "" {
        httpReq.Header.Set("X-Vault-AWS-IAM-Server-ID", authConfig.IAMServerID)
    }

    creds, err := stsOptions.Credentials.Retrieve(ctx)
    if err != nil {
        return nil, fmt.Errorf("failed to retrieve AWS credentials for STS request signing: %w", err)
    }
    if err := v4.NewSigner().SignHTTP(ctx, creds, httpReq, bodyHash, "sts", stsRegion, time.Now()); err != nil {
        return nil, fmt.Errorf("failed to sign STS GetCallerIdentity request: %w", err)
    }

    headers, err := json.Marshal(httpReq.Header)
    if err != nil {
        return nil, fmt.Errorf("failed to marshal signed STS request headers: %w", err)
    }

    return map[string]interface{}{
        "iam_http_request_method": http.MethodPost,
        "iam_request_url":         base64.StdEncoding.EncodeToString([]byte(stsEndpoint)),
        "iam_request_headers":     base64.StdEncoding.EncodeToString(headers),
        "iam_request_body":        base64.StdEncoding.EncodeToString([]byte(body)),
        "role":                    authConfig.Role,
    }, nil
}
```

internal/vault/client.go
```diff
@@ -198,6 +213,41 @@ func (c *Client) login(ctx context.Context) error {
 	return c.updateTokenMetadata(secret)
 }
 
+// resolvedSTSEndpointURL resolves the concrete STS endpoint using the SDK's
+// endpoint resolver, then normalizes it to a URL safe for signing.
+func resolvedSTSEndpointURL(ctx context.Context, stsSvc *sts.Client) (string, error) {
+	options := stsSvc.Options()
```
arjunkshashicorp: The function receives a `*sts.Client` but only calls `.Options()` on it — it never uses the client itself. Taking `sts.Options` directly would make the dependency more precise and easier to unit test in isolation:

```go
func resolveSTSEndpointURL(ctx context.Context, opts sts.Options) (string, error)
```

The caller already has `stsOptions` in scope so it's a one-line change at the call site.

internal/vault/client.go
```diff
@@ -126,64 +128,77 @@ func (c *Client) login(ctx context.Context) error {
...
+	d["iam_request_url"] = base64.StdEncoding.EncodeToString([]byte(stsEndpoint))
 	d["iam_request_headers"] = base64.StdEncoding.EncodeToString(headers)
-	d["iam_request_body"] = base64.StdEncoding.EncodeToString(body)
+	d["iam_request_body"] = base64.StdEncoding.EncodeToString([]byte(body))
 	d["role"] = c.authConfig.Role
 
 	secret, err := c.VaultClient.Logical().Write(fmt.Sprintf("auth/%s/login", c.authConfig.Provider), d)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to authenticate with Vault IAM auth provider %q: %w", c.authConfig.Provider, err)
```
arjunkshashicorp: Nit: `authConfig := c.authConfig` is used consistently throughout, but `c.authConfig.Provider` is accessed directly at the Write call. Since AuthConfig is a value type this makes no functional difference, but using the local `authConfig.Provider` here would be more consistent.

internal/proxy/proxy_test.go
```diff
@@ -63,11 +65,12 @@ var (
...
+	sts := ststest.FakeSTS(&awsCfg)
```
arjunkshashicorp: ### Name collision: `sts` variable shadows the imported package

```go
sts := ststest.FakeSTS(&awsCfg)
```

The variable name `sts` shadows the `sts` package if it were ever imported here. While it doesn't cause a bug today, it's a fragile pattern — a more descriptive name like `stsServer` or `fakeSTS` is safer and clearer, and is already used in `client_test.go`.

internal/proxy/proxy_test.go
```diff
@@ -4,6 +4,7 @@
 package proxy
```
arjunkshashicorp, line 4: ### `vaultRequests[1]` index in "happy path bare http client" is fragile

```go
require.Contains(t, vaultRequests[1].Header.Get("User-Agent"), proxyUserAgent)
```

This hardcodes the assumption that the login request is index `[0]` and the proxied request is `[1]`. If the login flow changes (e.g. token is already valid), this will panic with an index out of range. At minimum a `require.Len(t, vaultRequests, 2)` guard before accessing `[1]` would prevent a confusing failure.

arjunkshashicorp (changes requested):

internal/proxy/proxy_test.go
bcsamrudh: Addressed it in commit [57e9067](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/57e906776b6149f69eccc6fc0dcc8ab86c86b32c)

internal/proxy/proxy_test.go
bcsamrudh: Addressed it in commit [2a0cb0f](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/2a0cb0f33c97bb5d6c62ae25fa3fef8775795b42)

internal/vault/client.go
bcsamrudh: Addressed it in commit [c2880bd](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/c2880bd36e54db805ba2312f143754aa0d1e5f6d)

internal/vault/client.go
bcsamrudh: Addressed it in commit [c2880bd](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/c2880bd36e54db805ba2312f143754aa0d1e5f6d)

internal/vault/client.go
bcsamrudh: Addressed it in commit [c2880bd](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/c2880bd36e54db805ba2312f143754aa0d1e5f6d)

internal/vault/client.go
bcsamrudh: Addressed it in commit [c2880bd](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/c2880bd36e54db805ba2312f143754aa0d1e5f6d)

internal/vault/client.go
bcsamrudh: Addressed it in commit [c2880bd](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/c2880bd36e54db805ba2312f143754aa0d1e5f6d)

internal/vault/client.go
bcsamrudh: Addressed it in commit [c2880bd](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/c2880bd36e54db805ba2312f143754aa0d1e5f6d)

internal/vault/client.go
```diff
@@ -126,59 +128,72 @@ func (c *Client) login(ctx context.Context) error {
...
 
 	d := make(map[string]interface{})
-	d["iam_http_request_method"] = req.HTTPRequest.Method
-	d["iam_request_url"] = base64.StdEncoding.EncodeToString([]byte(req.HTTPRequest.URL.String()))
+	d["iam_http_request_method"] = http.MethodPost
+	d["iam_request_url"] = base64.StdEncoding.EncodeToString([]byte(stsEndpoint))
 	d["iam_request_headers"] = base64.StdEncoding.EncodeToString(headers)
-	d["iam_request_body"] = base64.StdEncoding.EncodeToString(body)
+	d["iam_request_body"] = base64.StdEncoding.EncodeToString([]byte(body))
 	d["role"] = c.authConfig.Role
```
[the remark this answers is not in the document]
bcsamrudh: Outdated

internal/vault/client.go
```diff
@@ -126,59 +128,72 @@ func (c *Client) login(ctx context.Context) error {
...
+	creds, err := stsOptions.Credentials.Retrieve(ctx)
 	if err != nil {
 		return err
 	}
+	if err := v4.NewSigner().SignHTTP(ctx, creds, httpReq, bodyHash, "sts", stsRegion, time.Now()); err != nil {
+		return err
+	}
 
-	body, err := io.ReadAll(req.HTTPRequest.Body)
+	headers, err := json.Marshal(httpReq.Header)
```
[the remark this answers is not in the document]
bcsamrudh: The AWS V4 signer already uses `httpReq.URL.Host` for the signature, so the signed request is valid without explicitly marshalling Host.

arjunkshashicorp (dismissed):
LGTM!

internal/vault/client.go
```diff
@@ -115,87 +115,150 @@ func (c *Client) RevokeToken() {
 
 // login authenticates to Vault using IAM auth, and sets the client's token.
 func (c *Client) login(ctx context.Context) error {
-	authConfig := config.AuthConfigFromEnv()
+	authConfig := c.authConfig
```
stuti-sr, line 118: It is a behavior change. If any callers ot tests relied on env changes after client creation, they no longer apply. 
Suggestion:
1. document this in PR notes or changelog if externally observable
2. add a test proving Client uses constructor-provided authConfig, not ambient env

internal/vault/client.go
```diff
@@ -115,87 +115,150 @@ func (c *Client) RevokeToken() {
...
+		return fmt.Errorf("error reading token: %s", err)
+	}
+	c.VaultClient.SetToken(token)
+
+	return c.updateTokenMetadata(secret)
+}
+
+// buildIAMAuthPayload builds and signs a GetCallerIdentity request, then packages
+// it into the payload expected by Vault's AWS IAM auth login endpoint.
+func buildIAMAuthPayload(ctx context.Context, stsSvc *sts.Client, authConfig config.AuthConfig) (map[string]interface{}, error) {
```
stuti-sr, line 184: I don’t see targeted unit tests around:

1. test case of Authorization, and other headers eg: X-Amz-Date, X-Amz-Security-Token
2. test case of X-Vault-AWS-IAM-Server-ID
3. behavior when credentials provider is missing, nil credentials provider returns a meaningful error
4. behavior when assumed credentials are used

internal/vault/client.go
```diff
@@ -115,87 +115,150 @@ func (c *Client) RevokeToken() {
...
+		region = defaultSTSRegion
+	}
+
+	endpoint, err := resolver.ResolveEndpoint(ctx, sts.EndpointParameters{
+		Region:            aws.String(region),
+		UseDualStack:      aws.Bool(opts.EndpointOptions.UseDualStackEndpoint == aws.DualStackEndpointStateEnabled),
+		UseFIPS:           aws.Bool(opts.EndpointOptions.UseFIPSEndpoint == aws.FIPSEndpointStateEnabled),
+		Endpoint:          opts.BaseEndpoint,
+		UseGlobalEndpoint: aws.Bool(false),
+	})
```
stuti-sr, line 248: I don’t see targeted unit tests around:

1. chosen STS endpoint URL
4. default region fallback to us-east-1

internal/vault/client.go
```diff
@@ -115,87 +115,150 @@ func (c *Client) RevokeToken() {
...
+	if region == "" {
+		region = defaultSTSRegion
+	}
+
+	endpoint, err := resolver.ResolveEndpoint(ctx, sts.EndpointParameters{
+		Region:            aws.String(region),
+		UseDualStack:      aws.Bool(opts.EndpointOptions.UseDualStackEndpoint == aws.DualStackEndpointStateEnabled),
+		UseFIPS:           aws.Bool(opts.EndpointOptions.UseFIPSEndpoint == aws.FIPSEndpointStateEnabled),
+		Endpoint:          opts.BaseEndpoint,
+		UseGlobalEndpoint: aws.Bool(false),
```
stuti-sr: If previous behavior effectively used the SDK default, this hard-coded false could alter host selection in some environments. Prefer inheriting SDK behavior unless there is a documented reason to force regional STS. If the force is intentional, add a comment explaining why.

internal/ststest/sts.go
```diff
@@ -7,26 +7,20 @@ import (
...
-	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/credentials"
 )
 
-// FakeSTS creates a fake STS server, and configures the session passed in
+// FakeSTS creates a fake STS server, and configures the AWS config passed in
 // to talk to that server.
-func FakeSTS(ses *session.Session) *httptest.Server {
+func FakeSTS(cfg *aws.Config) *httptest.Server {
```
stuti-sr: It makes tests a bit more coupled and easier to misuse if the same config is reused across cases.
Suggestion: return a copied config or provide a helper like:
`func FakeSTS(cfg aws.Config) (*httptest.Server, aws.Config)`

internal/vault/client_test.go
```diff
@@ -45,8 +44,9 @@ var (
 func TestTokenRenewal(t *testing.T) {
 	vault := fakeVault()
 	defer vault.Close()
-	ses := session.Must(session.NewSession())
-	stsServer := ststest.FakeSTS(ses)
+	awsCfg, err := awsconfig.LoadDefaultConfig(context.TODO())
```
stuti-sr, line 72: avoid using context.TODO() 
suggestion: context.Background()

internal/proxy/proxy_test.go
```diff
@@ -77,6 +80,7 @@ func TestProxy(t *testing.T) {
 
 		// the stored request should be the one _from the proxy_ since it's stored by
 		// the (fake) vault.
+		require.Len(t, vaultRequests, 2)
```
stuti-sr, line 84: provide better assertion: 
suggestion :
eg: 
// request 0 = login
assert.Contains(t, vaultRequests[0].URL.Path, "login")
// request 1 = proxied secret fetch
assert.Contains(t, vaultRequests[1].Header.Get("User-Agent"), proxyUserAgent)

bcsamrudh: Made change in [49a0e29](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/49a0e29f7d1c3b46121dcae0a245745091e3663c)

internal/vault/client.go
bcsamrudh: Made changes to stop forcing it to use regional endpoints (i.e UseGlobalEndpoint is no longer set to false)

internal/vault/client_test.go
bcsamrudh: Fixed it.

internal/vault/client.go
bcsamrudh: Added unit test cases targeting authorization and credential handling.

internal/ststest/sts.go
bcsamrudh: FakeSTS function now returns config also to prevent misuse.

internal/vault/client.go
bcsamrudh: added the testcases.

internal/vault/client.go
bcsamrudh: authConfig comes from main.go which gets the values from the env itself.https://github.com/hashicorp/vault-lambda-extension/blob/f6227900f03c3f91c336dcb4441b6e27c7c5b1a9/main.go#L115

internal/proxy/proxy_test.go
```diff
@@ -146,16 +154,21 @@ func TestProxy(t *testing.T) {
 	})
 }
 
-func startProxy(t *testing.T, vaultAddress string, ses *session.Session) (string, func() error) {
+func startProxy(t *testing.T, vaultAddress string, awsCfg aws.Config) (string, func() error) {
 	vaultConfig := api.DefaultConfig()
 	require.NoError(t, vaultConfig.Error)
 	vaultConfig.Address = vaultAddress
-	client, err := vault.NewClient("", "", hclog.NewNullLogger(), vaultConfig, config.AuthConfig{}, ses)
+	//client, err := vault.NewClient("", "", hclog.NewNullLogger(), vaultConfig, internalconfig.AuthConfig{}, awsCfg)
```
Balaji2198: Can we remove this commented-out code to keep the test clean?

internal/ststest/sts.go
```diff
@@ -7,26 +7,22 @@ import (
...
-			SessionToken:    "foo",
-		}))
+	cfg.BaseEndpoint = aws.String(stsServer.URL)
+	cfg.Region = "us-east-1"
+	cfg.Credentials = aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider("foo", "foo", "foo"))
 
-	return stsServer
+	cpycfg := *cfg
+
+	return stsServer, cpycfg
```
Balaji2198, line 26: I noticed `FakeSTS` mutates the original config and then returns a copy, which feels a bit confusing. Could we clarify the intent here, for example, can we either return the same config directly, or copy and update it up front, so the behavior is consistent?

internal/vault/client_test.go
```diff
@@ -361,6 +378,87 @@ func TestUserAgentHeaderAddition(t *testing.T) {
...
+	payload, err := buildIAMAuthPayload(context.Background(), stsSvc, config.AuthConfig{
+		IAMServerID: "vault.example.com",
+	})
+	require.NoError(t, err)
+
+	headers := decodeIAMRequestHeaders(t, payload)
+	require.Equal(t, "vault.example.com", headers.Get("X-Vault-AWS-IAM-Server-ID"))
+}
+
+func TestResolveSTSEndpointURL_DefaultsRegionToUSEast1(t *testing.T) {
```
Balaji2198, line 422: Can we add a testcase where both `authConfig.STSEndpointRegion` and `roleToAssumeArn` are set. This combination is a common production use-case for many customers

internal/proxy/proxy_test.go
```diff
@@ -158,7 +158,12 @@ func startProxy(t *testing.T, vaultAddress string, awsCfg aws.Config) (string, f
 	vaultConfig := api.DefaultConfig()
 	require.NoError(t, vaultConfig.Error)
 	vaultConfig.Address = vaultAddress
-	client, err := vault.NewClient("", "", hclog.NewNullLogger(), vaultConfig, internalconfig.AuthConfig{}, awsCfg)
+	//client, err := vault.NewClient("", "", hclog.NewNullLogger(), vaultConfig, internalconfig.AuthConfig{}, awsCfg)
```
arjunkshashicorp: please remove commented code if not needed.

arjunkshashicorp (changes requested):

internal/proxy/proxy_test.go
bcsamrudh: removed it

arjunkshashicorp (dismissed):
LGTM!

internal/proxy/proxy_test.go
```diff
@@ -146,16 +154,16 @@ func TestProxy(t *testing.T) {
 	})
 }
 
-func startProxy(t *testing.T, vaultAddress string, ses *session.Session) (string, func() error) {
+func startProxy(t *testing.T, vaultAddress string, awsCfg aws.Config) (string, func() error) {
 	vaultConfig := api.DefaultConfig()
 	require.NoError(t, vaultConfig.Error)
 	vaultConfig.Address = vaultAddress
-	client, err := vault.NewClient("", "", hclog.NewNullLogger(), vaultConfig, config.AuthConfig{}, ses)
+	client, err := vault.NewClient("", "", hclog.NewNullLogger(), vaultConfig, internalconfig.AuthConfig{}, awsCfg)
```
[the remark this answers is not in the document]
bcsamrudh: A non empty role is provided to reflect a realistic AuthConfig in startProxy.

Balaji2198 (commented):
Thanks for working on this. Added few comments.
For manual testing, could you please use the Quickstart example in [hashicorp/vault-lambda-extension](https://github.com/hashicorp/vault-lambda-extension/tree/main/quick-start) to verify these changes if that hasn’t been done already?

internal/proxy/proxy_test.go
bcsamrudh: already addressed in [4fb14f9](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/4fb14f99af89d87de4f873c104cf82c718bb38dd)

internal/ststest/sts.go
bcsamrudh: Updated the code in function FakeSTS to copy the config then modify it and return the copied config.

internal/vault/client_test.go
bcsamrudh: Added the testcase where both authConfig.STSEndpointRegion and roleToAssumeArn are set

bcsamrudh:
> For manual testing, could you please use the Quickstart example in [hashicorp/vault-lambda-extension](https://github.com/hashicorp/vault-lambda-extension/tree/main/quick-start) to verify these changes if that hasn’t been done already?

Thank you for the suggestion. I have already followed the Quickstart example and tested these changes on my end.

arjunkshashicorp (dismissed):
LGTM!

internal/proxy/proxy_test.go
```diff
@@ -63,11 +66,12 @@ var (
 func TestProxy(t *testing.T) {
 	fakeVault := fakeVault()
 	defer fakeVault.Close()
-	ses := session.Must(session.NewSession())
-	sts := ststest.FakeSTS(ses)
-	defer sts.Close()
-	proxyAddr, close := startProxy(t, fakeVault.URL, ses)
-	defer close()
+	awsCfg, err := awsconfig.LoadDefaultConfig(context.TODO())
```
stuti-sr: why are we using context.TODO() ?

internal/vault/client.go
```diff
@@ -115,87 +115,149 @@ func (c *Client) RevokeToken() {
...
+	stsOptions := stsSvc.Options()
+	if stsOptions.Credentials == nil {
+		return fmt.Errorf("failed to authenticate with Vault IAM auth provider %q: missing STS credentials provider", authConfig.Provider)
+	}
 
-	if c.authConfig.IAMServerID != "" {
-		req.HTTPRequest.Header.Add("X-Vault-AWS-IAM-Server-ID", c.authConfig.IAMServerID)
+	d, err := buildIAMAuthPayload(ctx, stsSvc, authConfig)
+	if err != nil {
+		return err
```
stuti-sr: any reason why error details are not added here?

internal/vault/client.go
```diff
@@ -115,87 +115,149 @@ func (c *Client) RevokeToken() {
...
+	httpReq, err := http.NewRequestWithContext(ctx, http.MethodPost, stsEndpoint, strings.NewReader(body))
 	if err != nil {
-		return err
+		return nil, fmt.Errorf("failed to build STS GetCallerIdentity request: %w", err)
 	}
-	if secret == nil {
-		return fmt.Errorf("got no response from the %s authentication provider", c.authConfig.Provider)
+	httpReq.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	if authConfig.IAMServerID != "" {
+		httpReq.Header.Set("X-Vault-AWS-IAM-Server-ID", authConfig.IAMServerID)
```
stuti-sr, line 205: can you please verify if v2 should use "github.com/aws/smithy-go/middleware" for IAM build phase and sign phase?

stuti-sr (changes requested):

internal/proxy/proxy_test.go
bcsamrudh: Fixed in [537ca60](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/537ca602f916147f3d93a4035fb3ef24ba5920eb)

internal/vault/client.go
bcsamrudh: Added them in [537ca60](https://github.com/hashicorp/vault-lambda-extension/pull/191/commits/537ca602f916147f3d93a4035fb3ef24ba5920eb)

internal/vault/client.go
bcsamrudh: No, "github.com/aws/smithy-go/middleware" isn't required as we are manually building a GetCallerIdentity request then signing it and sending it to Vault.

internal/vault/client_test.go
```diff
@@ -42,11 +51,27 @@ var (
...
+	return smithyendpoints.Endpoint{URI: *u}, nil
+}
+
 func TestTokenRenewal(t *testing.T) {
 	vault := fakeVault()
 	defer vault.Close()
-	ses := session.Must(session.NewSession())
-	stsServer := ststest.FakeSTS(ses)
+	awsCfg, err := awsconfig.LoadDefaultConfig(context.TODO())
+	require.NoError(t, err)
```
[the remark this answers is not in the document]
bcsamrudh: awsconfig.LoadDefaultConfig doesn't give error on a missing region, it just returns Region: "". 
Which is handled in client.go
https://github.com/hashicorp/vault-lambda-extension/blob/537ca602f916147f3d93a4035fb3ef24ba5920eb/internal/vault/client.go#L239-L241

arjunkshashicorp (dismissed):
LGTM!

stuti-sr (commented):
changelog is missing

internal/proxy/proxy_test.go
```diff
@@ -63,11 +66,12 @@ var (
...
 	fakeVault := fakeVault()
 	defer fakeVault.Close()
-	ses := session.Must(session.NewSession())
-	sts := ststest.FakeSTS(ses)
-	defer sts.Close()
-	proxyAddr, close := startProxy(t, fakeVault.URL, ses)
-	defer close()
+	awsCfg, err := awsconfig.LoadDefaultConfig(context.Background())
+	require.NoError(t, err, "failed to load AWS config ")
+	fakeSTS, awsCfg := ststest.FakeSTS(&awsCfg)
```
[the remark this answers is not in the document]
bcsamrudh: Not needed, as awsCfg is explicitly set in FakeSTS function in sts.go

https://github.com/hashicorp/vault-lambda-extension/blob/f6227900f03c3f91c336dcb4441b6e27c7c5b1a9/internal/ststest/sts.go#L21-L29

internal/vault/client_test.go
```diff
@@ -42,11 +51,27 @@ var (
...
+	stsServer, awsCfg := ststest.FakeSTS(&awsCfg)
```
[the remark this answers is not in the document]
bcsamrudh: Not needed, as awsCfg is explicitly set in FakeSTS function in sts.go

https://github.com/hashicorp/vault-lambda-extension/blob/f6227900f03c3f91c336dcb4441b6e27c7c5b1a9/internal/ststest/sts.go#L21-L29

arjunkshashicorp (dismissed):

Balaji2198 (dismissed):

stuti-sr (dismissed):
LGTM!!

stuti-sr (approved):
LGTM!!

arjunkshashicorp (approved):

Balaji2198 (approved):

PR #194, opened 2026-07-02
Fix STS Signature does not match non us-east-1

bcsamrudh:
Jira: https://hashicorp.atlassian.net/browse/VAULT-46705

Description:
Lambda function using vault-lambda-extension with AWS SDK V2 crashes when vault and lambda function are in a region other than `us-east-1` and also if vault and lambda function are in different regions.

Testing Evidence:

Note : `use_sts_region_from_client=false` i.e` VAULT_STS_ENDPOINT_REGION `is not used

**Case 1 : Vault and lambda in different region :** 

Vault is running in `ap-southeast-1` and Lambda Function is in `ap-southeast-2`

Creating Vault server in `ap-southeast-1` and lambda function in `ap-southeast-2` via terraform:
image

Result of lambda invoke:
image

This case was also tested using HCP where Vault cluster was in `us-west-2` and lambda function in `ap-south-1`

**Case 2: Both vault and lambda in same region other than us-east-1:**

Both the Vault and Lambda Function are in `eu-central-1`

Creating Vault server and Lambda in `eu-central-1` via terraform
image

Result of Lambda invoke
image

[VAULT-46705]: https://hashicorp.atlassian.net/browse/VAULT-46705?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/vault/client.go
```diff
@@ -183,14 +181,18 @@ func (c *Client) login(ctx context.Context) error {
...
-	if stsRegion == "" {
-		stsRegion = defaultSTSRegion
-	}
 
-	stsEndpoint, err := resolveSTSEndpointURL(ctx, stsOptions)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve STS endpoint URL: %w", err)
+	// Use the global STS endpoint (us-east-1) for the signed IAM login payload
+	// unless the caller explicitly configured a custom STS endpoint.
+	stsEndpoint, stsRegion := defaultSTSGlobalEndpoint, defaultSTSRegion
```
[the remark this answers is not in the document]
bcsamrudh: Resolved in [c50c834](https://github.com/hashicorp/vault-lambda-extension/pull/194/commits/c50c834d44bd5306e56074e5ee1c0df0c5d06efe)

internal/vault/client_test.go
```diff
@@ -509,7 +531,7 @@ func TestToken_UsesAssumedRoleArnWithSTSEndpointRegion(t *testing.T) {
 
 	payload := decodeVaultRequestPayload(t, vaultRequestBodies[0])
 	headers := decodeIAMRequestHeaders(t, payload)
-	require.Contains(t, headers.Get("Authorization"), "/eu-west-1/sts/")
+	require.Contains(t, headers.Get("Authorization"), "/us-east-1/sts/")
 }
```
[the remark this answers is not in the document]
bcsamrudh: Resolved in [c50c834](https://github.com/hashicorp/vault-lambda-extension/pull/194/commits/c50c834d44bd5306e56074e5ee1c0df0c5d06efe)

internal/vault/client.go
```diff
@@ -140,9 +141,6 @@ func (c *Client) login(ctx context.Context) error {
 		c.logger.Debug(fmt.Sprintf("Assumed role successfully with token expiration time: %s ", aws.ToTime(assumeRoleOutput.Credentials.Expiration).String()))
 
 		assumedRoleCfg := c.awsCfg.Copy()
-		if authConfig.STSEndpointRegion != "" {
-			assumedRoleCfg.Region = authConfig.STSEndpointRegion
-		}
 		assumedRoleCfg.Credentials = aws.NewCredentialsCache(credentials.NewStaticCredentialsProvider(
 			aws.ToString(assumeRoleOutput.Credentials.AccessKeyId),
```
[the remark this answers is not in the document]
bcsamrudh: Resolved in [c50c834](https://github.com/hashicorp/vault-lambda-extension/pull/194/commits/c50c834d44bd5306e56074e5ee1c0df0c5d06efe)

internal/vault/client.go
```diff
@@ -183,14 +181,18 @@ func (c *Client) login(ctx context.Context) error {
...
-		stsRegion = defaultSTSRegion
-	}
 
-	stsEndpoint, err := resolveSTSEndpointURL(ctx, stsOptions)
-	if err != nil {
-		return nil, fmt.Errorf("failed to resolve STS endpoint URL: %w", err)
+	// Use the global STS endpoint (us-east-1) for the signed IAM login payload
+	// unless the caller explicitly configured a custom STS endpoint.
+	stsEndpoint, stsRegion := defaultSTSGlobalEndpoint, defaultSTSRegion
+	if stsOptions.BaseEndpoint != nil {
```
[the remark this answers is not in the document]
bcsamrudh: Resolved in [c50c834](https://github.com/hashicorp/vault-lambda-extension/pull/194/commits/c50c834d44bd5306e56074e5ee1c0df0c5d06efe)

internal/vault/client.go
```diff
@@ -140,9 +142,6 @@ func (c *Client) login(ctx context.Context) error {
 		c.logger.Debug(fmt.Sprintf("Assumed role successfully with token expiration time: %s ", aws.ToTime(assumeRoleOutput.Credentials.Expiration).String()))
 
 		assumedRoleCfg := c.awsCfg.Copy()
-		if authConfig.STSEndpointRegion != "" {
-			assumedRoleCfg.Region = authConfig.STSEndpointRegion
-		}
```
stuti-sr, line 145: When assuming a role, the code creates a new AWS config via c.awsCfg.Copy() but only updates credentials. The critical STSEndpointRegion configuration is lost because it's not part of the base AWS config—it's stored separately in authConfig.STSEndpointRegion. This fine was critical.

internal/vault/client.go
stuti-sr, line 178: Region flows through three mechanisms: base awsCfg.Region, authConfig.STSEndpointRegion, and runtime override. Consolidate region handling at STS client creation. Also can you please provide comments ?

internal/vault/client_test.go
```diff
@@ -419,15 +419,38 @@ func TestBuildIAMAuthPayload_IncludesVaultIAMServerIDHeader(t *testing.T) {
 	require.Equal(t, "vault.example.com", headers.Get("X-Vault-AWS-IAM-Server-ID"))
 }
 
+func TestBuildIAMAuthPayload_DefaultPathUsesGlobalEndpointAndUSEast1Signing(t *testing.T) {
```
stuti-sr, line 422: Missing Test Coverage for Region Override

internal/vault/client.go
bcsamrudh: STSEndpointRegion Configuration isn't lost, its already applied in buildIAMAuthPayload function before signing
https://github.com/hashicorp/vault-lambda-extension/blob/d41864cdf6c61d2926cc15dbbb144e63b6e30ca4/internal/vault/client.go#L184-L186

stuti-sr (dismissed):
LGTM!

internal/vault/client_test.go
bcsamrudh: By "Missing Test Coverage for Region Override" if you mean to test the region override when STSEndpointRegion is provided. Its already covered in : https://github.com/hashicorp/vault-lambda-extension/blob/f0634e104fae21f397b1eb2965e769c7d26f59de/internal/vault/client_test.go#L445

internal/vault/client.go
```diff
@@ -241,13 +239,14 @@ func resolveSTSEndpointURL(ctx context.Context, opts sts.Options) (string, error
...
 	endpoint, err := resolver.ResolveEndpoint(ctx, sts.EndpointParameters{
-		Region:       aws.String(region),
-		UseDualStack: aws.Bool(opts.EndpointOptions.UseDualStackEndpoint == aws.DualStackEndpointStateEnabled),
-		UseFIPS:      aws.Bool(opts.EndpointOptions.UseFIPSEndpoint == aws.FIPSEndpointStateEnabled),
-		Endpoint:     opts.BaseEndpoint,
+		Region:            aws.String(region),
+		UseDualStack:      aws.Bool(opts.EndpointOptions.UseDualStackEndpoint == aws.DualStackEndpointStateEnabled),
+		UseFIPS:           aws.Bool(opts.EndpointOptions.UseFIPSEndpoint == aws.FIPSEndpointStateEnabled),
+		Endpoint:          opts.BaseEndpoint,
+		UseGlobalEndpoint: aws.Bool(opts.BaseEndpoint == nil),
```
arjunkshashicorp: `UseGlobalEndpoint` is conditional only on `BaseEndpoint`:

```go
UseGlobalEndpoint: aws.Bool(opts.BaseEndpoint == nil),
```

In production, `main.go` applies `VAULT_STS_ENDPOINT_REGION` only as `awsconfig.WithRegion(...)` — it never sets `BaseEndpoint`. So `opts.BaseEndpoint` is always nil, `UseGlobalEndpoint=true`, and the SDK endpoint resolver ignores `opts.Region` entirely, always returning `sts.amazonaws.com` signed with `us-east-1`. Setting `VAULT_STS_ENDPOINT_REGION=eu-west-1` has no effect on the endpoint URL or signing region.

This breaks users whose Vault IAM auth mount uses a regional STS endpoint (e.g. `sts_endpoint=https://sts.eu-west-1.amazonaws.com`). Their requests will be signed for the wrong endpoint and region, causing auth failure.

Note: `TestToken_UsesAssumedRoleArnWithSTSEndpointRegion` doesn't catch this because it sets `BaseEndpoint: aws.String(stsServer.URL)`, which makes `UseGlobalEndpoint=false` in the test — the opposite of what happens in production.

It would be good to add a test for `buildIAMAuthPayload` that sets `STSEndpointRegion` but no `BaseEndpoint` (matching real production config), and asserts that the `Authorization` header contains the expected region and the `iam_request_url` points to the regional STS endpoint.

**Suggested fix** — decouple the global-endpoint decision from `STSEndpointRegion`:

```go
func buildIAMAuthPayload(ctx context.Context, stsSvc *sts.Client, authConfig config.AuthConfig) (map[string]interface{}, error) {
    opts := stsSvc.Options()
    useGlobalEndpoint := opts.BaseEndpoint == nil
    if authConfig.STSEndpointRegion != "" {
        opts.Region = authConfig.STSEndpointRegion
        useGlobalEndpoint = false
    }
    stsEndpoint, stsRegion, err := resolveSTSEndpoint(ctx, opts, useGlobalEndpoint)
```

```go
func resolveSTSEndpoint(ctx context.Context, opts sts.Options, useGlobalEndpoint bool) (endpointURL, signingRegion string, err error) {
    ...
    endpoint, err := resolver.ResolveEndpoint(ctx, sts.EndpointParameters{
        ...
        UseGlobalEndpoint: aws.Bool(useGlobalEndpoint),
    })
```

arjunkshashicorp (changes requested):

internal/vault/client.go
suraj-simha: good catch, thank you.
Updated the code by decoupling the `useGlobalEndpoint` from `STSEndpointRegion`

internal/vault/client.go
```diff
@@ -182,13 +181,18 @@ func (c *Client) login(ctx context.Context) error {
 // buildIAMAuthPayload builds and signs a GetCallerIdentity request, then packages
 // it into the payload expected by Vault's AWS IAM auth login endpoint.
 func buildIAMAuthPayload(ctx context.Context, stsSvc *sts.Client, authConfig config.AuthConfig) (map[string]interface{}, error) {
-	stsOptions := stsSvc.Options()
-	stsRegion := stsOptions.Region
-	if stsRegion == "" {
-		stsRegion = defaultSTSRegion
+	opts := stsSvc.Options()
```
stuti-sr, line 184: stsSvc.Options() returns a copy of the options struct. Modifying opts.Region only affects the local copy
This can causes a region mismatch between the signed request and the actual endpoint. Can you provide test results for : 
1. Lambda functions when deployed outside us-east-1
2. Cross-region Vault authentication
3. Test in multi-region deployments

internal/vault/client.go
```diff
@@ -182,13 +181,18 @@ func (c *Client) login(ctx context.Context) error {
 // buildIAMAuthPayload builds and signs a GetCallerIdentity request, then packages
 // it into the payload expected by Vault's AWS IAM auth login endpoint.
 func buildIAMAuthPayload(ctx context.Context, stsSvc *sts.Client, authConfig config.AuthConfig) (map[string]interface{}, error) {
```
stuti-sr: Since this is performing critical region resolution , can we have better logging with: c.logger.Debug("resolved STS endpoint","endpoint_url", stsEndpoint,"signing_region", stsRegion)

internal/vault/client.go
```diff
@@ -182,13 +181,18 @@ func (c *Client) login(ctx context.Context) error {
...
-	stsOptions := stsSvc.Options()
-	stsRegion := stsOptions.Region
-	if stsRegion == "" {
-		stsRegion = defaultSTSRegion
+	opts := stsSvc.Options()
+	// Default to the global STS endpoint when no custom endpoint is configured.
+	useGlobalEndpoint := opts.BaseEndpoint == nil
```
stuti-sr: nit : 
 “use SDK default endpoint resolution, `useDefaultResolver`” not necessarily “always use global. `useGlobalEndpoint`”

stuti-sr (commented):
provided comments. Can you please add jira story in description?

internal/vault/client.go
suraj-simha: Good observation. 
It doesn't cause a region mismatch because that same local, mutated `opts` copy is threaded through both `resolveSTSEndpoint` (endpoint resolution) and the SigV4 signing call later in `buildIAMAuthPayload`, so the endpoint URL and the signing region always stay consistent with each other. `stsSvc` is never re-queried after the mutation, so the fact that its own state is unchanged is harmless.

Attached the test result for the same in the PR description.
Also the tests `TestBuildIAMAuthPayload_DefaultPathUsesGlobalEndpointAndUSEast1Signing` and `TestToken_UsesAssumedRoleArnWithSTSEndpointRegion` are having the same behaviour mentioned.

internal/vault/client.go
suraj-simha: updated, thanks.

internal/vault/client.go
suraj-simha: updated `useGlobalEndpoint` to `useDefaultResolver`.

arjunkshashicorp (dismissed):
LGTM!

stuti-sr (commented):
LGTM!!

stuti-sr (dismissed):

santoshhashicorp (dismissed):
LGTM

internal/vault/client_test.go
```diff
@@ -512,6 +535,64 @@ func TestToken_UsesAssumedRoleArnWithSTSEndpointRegion(t *testing.T) {
 	require.Contains(t, headers.Get("Authorization"), "/eu-west-1/sts/")
 }
 
+// TestToken_RegionalSTSEndpointWithoutBaseEndpoint verifies that when STSEndpointRegion
+// is set without BaseEndpoint (matching production config), the regional endpoint is used
+// and requests are signed for the correct region.
+func TestToken_RegionalSTSEndpointWithoutBaseEndpoint(t *testing.T) {
```
Balaji2198, line 612: Can we add a test case to cover the behavior when `opts.BaseEndpoint` is set? It'd be great to verify that custom endpoints resolve and sign correctly.

internal/vault/client.go
```diff
@@ -257,7 +269,16 @@ func resolveSTSEndpointURL(ctx context.Context, opts sts.Options) (string, error
 		u.Path = "/"
 	}
 
-	return u.String(), nil
+	signingRegion = region
+	if authOpts, ok := smithyauth.GetAuthOptions(&endpoint.Properties); ok {
```
Balaji2198, line 273: Can we add an unit test where `GetAuthOptions` returns false when `endpoint.Properties` is missing? This will help ensure the `signingRegion` fallback to `opts.Region` works as expected and won't break in the future.

Balaji2198 (commented):
Thanks for working on this. Provided few comments.

internal/vault/client_test.go
suraj-simha: Added `TestBuildIAMAuthPayload_CustomBaseEndpointIsUsedForURLAndSigning` + `TestResolveSTSEndpoint_CustomBaseEndpoint` covering the scenario.
Thank!

internal/vault/client.go
suraj-simha: Thanks! Added `TestResolveSTSEndpoint_FallsBackToOptsRegionWhenAuthOptionsMissing` for the above scenario.

Balaji2198 (approved):

